PR #4, opened 2020-02-17
Added media content and favicon

franciscop:
Added content for sharing the image in different media (Twitter, Facebook, Whatsapp, etc). Recommended to update screenshot semi-regularly.

Using Japan map now as favicon:

favicon-32x32

Could switch to microbe if preferred (though it seemed a bit extreme/scary):

favicon-32x32


reustle:
Thoughts on using the Japanese flag as the favicon?

image

franciscop:
The Japanese flas seems like a recording button in there. AFAIK I've seen some browsers use a red circle to indicate there's a video feed in the tab.

I normally keep the lossless png since the jpg is lossy, so any future modification happen on the png and are exported to jpg without losing quality. Can delete if not needed at all.

Issue #7, opened 2020-02-18
Japanese translation (first phase)


khasunuma:
@reustle I've quickly translate `index.html` to Japanese and save it as `index_ja.html` in my repository forked from your original. I'd like to contribute the site as a Japanese person.

I'll send you the pull request after a few minutes. Since the map doesn't work in my local environment, Could you please modify to work well if you would accept the pull request.

I want to translate prefectures' name in the "Prefecture Data" table, however, it's difficult to translate quickly for now. It's my challenge.

reustle:
For those checking on this issue, I am replying on the pull request

PR #8, opened 2020-02-18
Add Japanese translation (in progress)

khasunuma:
As I told the issue #7, I'd like to contribute first Japanese translation of this site. It's not complete translation because the contents of "Prefecture Data" (especially prefectures' name) has not translated yet, and it's difficult to translate quickly for now. It's my challenge and hope you provide full translation if possible.

### DIFF
--- a/docs/index_ja.html
+++ b/docs/index_ja.html
@@ -1,0 +1,166 @@
+<!DOCTYPE HTML>
+<html lang="ja">
+<head>
+
+    <!-- Warning: Ugly code lies ahead -->
+
+
+    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1, minimal-ui">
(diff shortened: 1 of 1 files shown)


reustle:
Thank you @khasunuma san for doing this! I am currently researching how to translate the rest of the page. I will get back to you here shortly.

Changing (some of) the map labels:
```
map.setLayoutProperty('country-label', 'text-field', ['get','name_ja']);
```

reustle:
An update, I have still not come up with a nice solution to translate the prefecture names smoothly. I may just create a manually list and switch between them, but that is less than ideal. I will spend a little bit more time on the idea and then will implement something, whether it is the best way or not. Thank you again for your work here!

reustle:
Hello @khasunuma san, I have integrated your changes in the other PR and deployed it. It is not entirely 100% translated yet, but some of the content is quite difficult to translate in a nice way. I hope to receive / create a PR soon to properly translate the trend chart, as it was having some issues with my attempts to swap text.

Thanks again!

PR #21, opened 2020-02-22
Japanese Language Support

reustle:
@khasunuma san has gone through the entire page and translated everything in to Japanese with his PR #8 

Seeing that there isn't much text, we feel it makes sense to handle translation switching entirely on the client-side. We can even set a url hash like `/#ja` to remember the language selection.

This PR is the first pass at catching all English and making sure it is translated.

Translation function by @franciscop

### DIFF
--- a/index.html
+++ b/index.html
@@ -158,6 +158,24 @@
   <script src="https://cdn.jsdelivr.net/npm/lodash@4.17.15/lodash.min.js"></script>
   <script src="https://cdn.jsdelivr.net/npm/drive-db@5.0.3/index.min.js"></script>
   <script src="static/main.js"></script>
+  <script>
+    (function(){
+      initDataTranslate()
+      drawMap()
+
+      // TODO This should load the data in parallel
+      map.once('style.load', function(e) {
+        loadPrefectureData(function(){
+          drawMapPrefectures(ddb.prefectures)
+          drawPrefectureTable(ddb.prefectures)
+        })
+      })
+
+      loadTrendData(function() {
+        drawTrendChart(ddb.trendData)
+      })
+    })()
+  </script>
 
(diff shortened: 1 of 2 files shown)


reustle:
Current progress is shown below. Still need to do the dates on the chart, update the language selector, and clean up the switching code. (cc #8)

image

index.html
```diff
@@ -55,31 +60,31 @@ <h1>Coronavirus Disease (COVID-19) Japan Tracker</h1>
...
+        <td data-ja="回復者数">Recovered</td>
+        <td data-ja="死亡者数">Deceased</td>
       </tr>
     </table>
   </section>
 
   <a name="map"></a>
   <div class="center-offset">
-    <h4>Cases By Prefecture</h4>
+    <h4 data-ja="都道府県ごとの件数">Cases By Prefecture</h4>
```
Kay-Cube: The Japanese is the direct translation from the English.
But here I think the English name should also change a bit, so it will differ from the "Prefecture Data"
Can we call it "Outbreak Map"? 
"ウイルス感染マップ”

index.html
```diff
@@ -90,14 +95,14 @@ <h4>Outbreak Spread Trend</h4>
...
       <thead>
         <tr>
-          <th>Prefecture</th>
-          <th>Confirmed</th>
-          <th>Recovered</th>
-          <th>Deceased</th>
+          <th data-ja="都道府県">Prefecture</th>
+          <th data-ja="感染者数">Confirmed</th>
+          <th data-ja="回復者数">Recovered</th>
+          <th data-ja="死亡者数">Deceased</th>
```
Kay-Cube: Don't know where to add, but "Total"'s japanese is ”計”

reustle: Added

index.html
reustle: Updated

PR #23, opened 2020-02-23
Adding 'Delta' series with rate-of-change of confirmed cases

noodlefrenzy:
I really want to get a sense of the first and second derivative of the confirmed cases, so I can differentiate between 'mild alarm', 'existential dread', and 'full-blown panic'. 'Delta' is kind of a wonky name, so let me know if you'd like me to change that, and I chose the color pretty much at random so let me know if there was a palette you were pulling from.


reustle:
Great, thanks for submitting this! Could you attach a screenshot as well please? @Kay-Cube will know about coloring. Also I'm assuming the KPIs are now 4x wide, not sure if we want to go to 2 rows yet (mobile only?)

Cheers

noodlefrenzy:
> Great, thanks for submitting this! Could you attach a screenshot as well please? @Kay-Cube will know about coloring. Also I'm assuming the KPIs are now 4x wide, not sure if we want to go to 2 rows yet (mobile only?)
> 
> Cheers

Actually I didn't add it to the KPIs, just the graph, since a singular "delta" value is not very useful and I worried about the page-flow (as you mentioned). Let me know if you disagree.

noodlefrenzy:
image

Here's a shot of the revised graph and legend.

noodlefrenzy:
I've merged upstream (sorry, took a while to install jekyll and got distracted by my 'job' and 'friends' 😅). Updated naming. I'd love to merge this in and work on toggling afterwards since I think that's a separate feature (and am hoping for a response on why we're not using the native legend which has that built-in). Let me know.

noodlefrenzy:
Oh, P.S. my data-js entry for 'Daily Increase' was done by Google Translate (勉強しますけどちょっとばかです). Please correct me if I'm wrong.

static/main.js
```diff
@@ -10,6 +10,7 @@ const TIME_FORMAT = 'YYYY-MM-DD'
 const COLOR_CONFIRMED = 'rgb(244,67,54)'
 const COLOR_RECOVERED = 'rgb(25,118,210)'
 const COLOR_DECEASED = 'rgb(55,71,79)'
+const COLOR_INCREASE = 'rgb(123,71,19)'
```
reustle:
```suggestion
const COLOR_INCREASE = 'rgb(163,172,191)'
```
Updated color to grey

noodlefrenzy:
That's what I thought re: reason on legend. I've updated the color, please let me know if you want me to do anything further before you merge. Once it's merged, I'll start working on some toggle logic - let me know if you'd like to be able to toggle all datasets, or just this one.

static/css/styles.scss
```diff
@@ -153,6 +153,9 @@ h1 {
   span.three {
     color: rgb(55,71,79);
   }
+  span.four {
+    color: rgb(123,71,19);
```
reustle:
```suggestion
    color: rgb(163,172,191);
```
oops missed this

PR #43, opened 2020-02-26
Get data action

olingern:
This should work if the correct environment variables are provided. I had to duplicate the `cache-sheet` file and the dependencies due to Docker permissions and not being able to reach above the current directory.  
image

Github notes that 5 minutes is the minimum for cron schedules, but I'm seeing 8 minutes. I guess they're still working things out


olingern:
Also, it looks like requesting a reviewer permissions aren't set up. I'm unable to select a user from the "Reviewers" area

image

reustle:
Thanks so much for this @olingern ! 

We're (@python273 and I) are going through this now, but a bit delayed because [Github has been down](https://www.githubstatus.com/) 😅 

Will comment here once confirmed/tested.

reustle:
Merged! Thanks a ton Nick @olingern - You can see it in _action_ here: https://github.com/reustle/covid19japan/actions

olingern:
@reustle welcome! :rocket:

Issue #27, opened 2020-02-25
Add City-Level Data


reustle:
Do we have the ability to move from prefecture-level to city-level? It seems like many published cases include this information.

How would they be visualized?

amwfarid:
Assuming the data can be made available on the spreadsheet, I can propose the following implementations:

**Map perspective:**
We can visualize city level statistics by plotting lat/lng markers (of cities) on the map once a close-enough zoom level is reached (say at the present max zoom=7).

The current geojson only contains the bounds for the prefectures, it'd be time consuming to track down geojson files per each prefecture and integrating them. I haven't yet found one geojson file that has city level separation on the entire level of Japan.


**Graph perspective:**
We could add two drop menus (prefecture -> city) in the "spread trend" area and reload the graph accordingly.

exhaze:
This is the raw data published by MLIT: http://nlftp.mlit.go.jp/ksj/gml/datalist/KsjTmplt-N03-v2_3.html

It has GeoJSON data for every single city and prefecture.

reustle:
Thinking we should use Clustering to show this data.

https://docs.mapbox.com/mapbox-gl-js/example/cluster/

https://blog.mapbox.com/clustering-properties-with-mapbox-and-html-markers-bb353c8662ba

Question I still would like to hear suggestions about, how do we handle cases with no city defined? Do we default them to the center of the prefecture? The biggest city of the prefecture?

![image](https://user-images.githubusercontent.com/304560/75733512-e03f0680-5d38-11ea-827f-19b076bc6bb4.png)

![image](https://miro.medium.com/max/862/0*g09phU_4zwOEcd-8.gif)

liquidx:
That's very cool. 

My suggestion is for the cases we don't have the data, center of the prefecture is probably good enough since it could be any of the towns, not necessarily the biggest one.

l15n:
> My suggestion is for the cases we don't have the data, center of the prefecture is probably good enough since it could be any of the towns, not necessarily the biggest one.

That's what https://coromap.info/ does, so having a look at the cases can give you a good idea of what it looks like. e.g. Case 117's visit to Hiroshima Prefecture, Case 328 in Okayama etc.

olingern:
I'm interested in this as I'm a semi immune compromised person who is trying to find the best place to wait the pandemic out.

[...]

Issue #39, opened 2020-02-26
Static Map Image


reustle:
Instead of loading mapbox on every page load, let's generate an image of the map and prefectures once every 5 minutes (like we do with the json cache) and store it on s3.

Once the user clicks the image, it will then load the real mapbox which they can interact with. I would bet that >75% of users don't touch the map, so loading mapbox isn't necessary.

python273:
I think we could take geojson and convert to svg to replace mapbox completely

https://github.com/deldersveld/topojson/blob/master/countries/japan/jp-prefectures.json
https://exploratory.io/map

reustle:
This is a great idea, I'm wondering how it will change the way we do #27. Since we seem to have geojson data there too, it may not be a problem?

python273:
https://exploratory.io/map has cities dataset, but no English names and each city is not a single ploygone
<img width="1015" alt="Screenshot 2020-02-27 at 17 04 14" src="https://user-images.githubusercontent.com/3097956/75452509-7f53af00-5983-11ea-8ea6-1ed4ea004159.png">

python273:
this one seems slightly better
https://github.com/deldersveld/topojson/blob/master/countries/japan/jp-towns.json

reustle:
FTR

https://docs.mapbox.com/help/how-mapbox-works/static-maps/

https://api.mapbox.com/styles/v1/mapbox/light-v10/static/139.083,38.0965,4.03/650x400?access_token=pk.eyJ1IjoicmV1c3RsZSIsImEiOiJESzd6YVRnIn0.Hh9AwQw1X0PR_TOewZMMzA

olingern:
From #47 

> Due to the quickly rising cost of hosting Mapbox on the site

I'm wondering if instead of switching to svg from mapbox, someone reaches out to mapbox and see if they would give the project a pass since it's benefiting many people in Japan

reustle:
https://twitter.com/reustle/status/1234371319321387008 🤞

reustle:
They've offered to help out which is great, but I still think something like their 'static map' on load would be nice for the user. Once you click the map, it loads the interactive version in its place.

PR #47, opened 2020-02-26
implement city hover and small refeactor

### LINKED ISSUE #32: Show Prefecture Case Count On Hover
When you hover over a prefecture on the map, show the number of cases

olingern:
Fixes #32.

Builds a lookup from s3 json and then performs lookups on mouse hovers via the geo json's city.

image

I saw that `.map` was being used in a couple of places, but the returned array wasn't being used. Changed those to for loops.

### DIFF
--- a/static/main.js
+++ b/static/main.js
@@ -22,38 +22,41 @@ let ddb = {
   }
 }
 let map = undefined
+const prefectureCaseLookup = {}
 
 
 function loadData(callback) {
   // Load the json data file
-  
+
   fetch(JSON_PATH).then(function(res){
     return res.json()
   })
   .then(function(data){
     callback(data)
-  })
+    })
 }
 
 
 function calculateTotals(prefectures) {
   // Calculate the totals
-  
-  let totals = {
+
+  const totals = {
     confirmed: 0,
     recovered: 0,
     deceased: 0,
   }
-  
-  prefectures.map(function(pref){
+
+  for (let i = 0; i < prefectures.length; i++) {
+    const pref = prefectures[i];
+    prefectureCaseLookup[pref.prefecture] = pref.cases && pref.cases > 0 ? pref.cases : 0;
+
     // TODO change to confirmed
-    totals.confirmed += (pref.cases?parseInt(pref.cases):0)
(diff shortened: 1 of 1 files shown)


amwfarid:
@olingern You mentioned in the title "city hover". Would it also happen to resolve #27 as well?

reustle:
We don't quite have city data logged yet in the spreadsheet, but will be working on it. I added a column for it last night, so we will start to see it being added.

reustle:
Great, thanks for implementing this! Here are a few comments.

1. Do you know if a simple way to center the pin on the prefecture always? Would we need to have a defined `center point` for each prefecture, or can we reasonably easily calculate it using prefectures json?

2. I think the label should say something like `12 confirmed cases`

3. Can we somehow not show any popup on prefectures with no cases?

4. Due to the quickly rising cost of hosting Mapbox on the site, #39 is looking like something that will be done quite soon. It might be better to hold off on merging this until that is done? Or we quickly merge it and let it run until we implement the new map. More of an FYI

Only numbers 2 & 3 above are important. Thanks again!

image

reustle:
Ah, sooner than expected https://github.com/reustle/covid19japan/pull/49

olingern:
1. Could be tricky. Maybe it's possible to sort by lat / lng, but doing so doesn't seem super performant. Arbitrarily selecting an index, _i.e. always 0_ would probably work best. I think the geojson would need the center point as a property if desired.

2. Sure, but we would need Japanese, which I'm only mediocre at. Maybe "12 確認しました"  ... [NHK](https://www3.nhk.or.jp/news/html/20200227/k10012303921000.html) seems to be using "感染確認" though. So, in their style, "感染確認12人", but I'm sure any Japanese person would know best.

3. Yah, doable.

4. I see. I can hold off until #49 is done and rebase that over this or close and re-open a new PR.

reustle:
4. 👌 I think this is the best choice, let's jump right back on this as soon as SVG is merged. Thanks again!

olingern:
Closing this for now,

reustle:
Thank you again! We're not so sure about the future of the map yet overall.

olingern:
No prob. I'll try to drop in again and offer some help somewhere else :slightly_smiling_face:

PR #49, opened 2020-02-27
Switch from mapbox to static SVG map

liquidx:
@reustle 

I replaced the MapBox map with a static SVG map with prefecture level highlighting. Whatcha think?

Screen Shot 2020-02-27 at 22 54 24

The map is licensed CC-BY-NC.


static/main.js
```diff
@@ -276,68 +244,38 @@ function drawPageTitleCount(confirmed) {
...
-  })
 
-  // Start the Mapbox search expression
-  let prefecturePaint = [
-    'match',
-    ['get', 'NAME_1'],
-  ]
-  
   // Go through all prefectures looking for cases
+  let japanMapSvg = document.querySelector(`#map-japan-svg`).contentDocument
```
reustle: \` isn't supported in IE unfortunately

python273:
I'd also crop svg, so it's zoomed in a bit more

liquidx:
Let me take a look and try this evening to make it work across mobile and IE.

Zooming should be doable, I think I should add some labels which is also possible.

static/main.js
liquidx: Fixed

liquidx:
Fixed layout on mobile, but didn't add zooming.

Looks like the site is still broken on IE somewhat, even before getting to drawing the map.

olingern:
ping @reustle, looks like this is waiting for review / comment

reustle:
Sorry, @liquidx and I talk offline a lot. Currently wrangling how mobile will work, and how we would then move to city-based data instead of prefecture-based. Kind of 'on-hold'.

reustle:
Do we want to shelve this PR for the time being?

olingern:
I think so. Will close :+1:

liquidx:
Closing this PR. HEAD has deviated so far and we ended up being OK with the MapBox map that I think it's fine to leave it as is for now.

Issue #40, opened 2020-02-26
Bundle JS Libraries


reustle:
We are importing a lot of libraries at this point. Let's pull them all into a single file and save them within the repo, and load them as one file. Probably good to use `webpack`.

tatellos:
Will do this in conjunction with #41 .

olingern:
Just wanted to comment and say that bundling, IE support, pwa, front-end framework, etc. is handled out of the box in [Gatsby](https://www.gatsbyjs.org/) (another static site generator). Not that the project should switch, but re-inventing an existing tool chain is something to be aware of.

reustle:
Good to know, thanks @olingern - One of the reasons I pick Jekyll is because github pages supports it out of the box. Looks like [using gatsby with github pages](https://www.gatsbyjs.org/docs/how-gatsby-works-with-github-pages/#deploying-to-a-path-on-github-pages) still requires running some build commands. Not the end of the world, but worth noting.

PR #51, opened 2020-02-27
JS Bundling and Pollyfills

### LINKED ISSUE #41: Add FETCH Polyfill
`fetch` needs a polyfill on older browsers

tatellos:
Fixes #41 
Create package.json, add webpack config, create new src folder.
/static should become our output folder, i.e. the other non-generated files should move from there to /src as well.
We probably won't need jekyll anymore then, but I'm not sure how the site is served on prod, so I didn't change that concept.

// Please assign @reustle to your newly created PR

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 _site/
 
 node_modules
+.jekyll-cache/
--- a/embed.html
+++ b/embed.html
@@ -37,10 +37,7 @@ layout: base
 <!-- JS -->
 <script src="https://api.mapbox.com/mapbox-gl-js/v0.53.1/mapbox-gl.js"></script>
 <link href="https://api.mapbox.com/mapbox-gl-js/v0.53.1/mapbox-gl.css" rel="stylesheet" />
-<script src="https://cdn.jsdelivr.net/npm/moment@2.24.0/moment.min.js"></script>
-<script src="https://cdn.jsdelivr.net/npm/chart.js@2.8.0"></script>
-<script src="https://cdn.jsdelivr.net/npm/lodash@4.17.15/lodash.min.js"></script>
-<script src="https://cdn.jsdelivr.net/npm/fetch-polyfill@0.8.2/fetch.min.js"></script>
+<script src="https://cdn.jsdelivr.net/combine/npm/moment@2.24.0/moment.min.js,npm/chart.js@2.8.0,npm/lodash@4.17.15/lodash.min.js,npm/fetch-polyfill@0.8.2/fetch.min.js"></script>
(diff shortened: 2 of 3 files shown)


tatellos:
@reustle 
I'm still getting errors in IE11, one of them directly from the drive-db file we pull in.
They might go away when I'm finished with including modules through webpack (#40 ). I'll also use Babel and compile to ES5, so that the output code is compatible with older browsers...

tatellos:
I forgot embed.html.
I'll refactor that to use the same JavaScript as index.html.

reustle:
Hey Tatellos!

Our goal with this is actually to reduce the number of 3rd party libraries down into a single file, sorry that it wasn't more well defined in the issue. I don't think it's necessary to webpack our single main.js file or the code embedded in the page, as that would complicate the build process.

Ideally, this would take all 3rd party decency scripts and put them into a single file, I think usually called vendor.js? Then we'll continue to import main.js as usual.

Please let me know your thoughts, thanks!

tatellos:
I see. But when you are saying we want to reduce the third party libraries into a single file, that is a build process.
Not including the main script in that build process would actually be more complicated than including it.

Using webpack to compile the whole project (including main.js) would also make it easier to 
* separate out main.js into different files so that it becomes more maintainable. 
* enable modern JavaScript syntax, e.g. use arrow functions again, but have the output still work in IE (I was going to do this while fixing #40 ).

etc.

If you really only want Jekyll as a dependency, you will be struggling a lot with JavaScript. I don't recommend that.

python273:
@tatellos do you know if there's some way to build JS before it's getting deployed to github pages?

Maybe it would be simpler to just replace `fetch` with `XMLHttpRequest`. And there's no that much js code, so might be fine to use "non-modern" js 😅

reustle:
It looks like jsDelivr, which is where we pull almost all resources, supports combining multiple files into a single request https://www.jsdelivr.com/features

```
https://cdn.jsdelivr.net/combine/url1,url2,url3
```
```
https://cdn.jsdelivr.net/combine/gh/jquery/jquery@3.2/dist/jquery.min.js,gh/twbs/bootstrap@3.3/dist/js/bootstrap.min.js
```

tatellos:
@python273 There doesn't seem to be a good way to run webpack for github pages. I didn't know that that's how we're serving prod.

So I just combined the files into one request, and removed the rest (webpack etc)

tatellos:
@python273 So I just discovered Github Workflows ^^ 
Just for future's sake, I'm pretty sure that that would be a good place to do the webpack bundling

PR #64, opened 2020-03-02
implement autoreload data

### LINKED ISSUE #10: Auto Reload Data
Automatically reload the data source after 5 minutes. Once reloaded, if there are changes, redraw the elements on the page and the page title.

We don't need to re-draw the map, just the data layer.

olingern:
@reustle 

Fixes #10 

edit:

**To test**:
- Check out branch
- Change `FIVE_MINUTES_IN_MS` to something lower res, i.e. 3000
- Load the page and inspect the network tab

### DIFF
--- a/index.html
+++ b/index.html
@@ -145,44 +145,60 @@ layout: base
     if(typeof drawMap === 'undefined'){
       return
     }
-    
+
     initDataTranslate()
     drawMap()
-    
+
+    var pageDraws = 0;
     var styleLoaded = false
     var jsonData = undefined
-    
+    const FIVE_MINUTES_IN_MS = 300000
+
     function whenMapAndDataReady(){
       if(!styleLoaded || !jsonData){
         return
       }
-      
-      drawMapPrefectures(ddb.prefectures)
+
+      drawMapPrefectures(pageDraws)
     }
-    
+
     map.once('style.load', function(e) {
       styleLoaded = true
       whenMapAndDataReady()
     })
-    
-    loadData(function(data){
-      jsonData = data
-      
-      ddb.prefectures = jsonData.prefectures
-      let newTotals = calculateTotals(jsonData.daily)
-      ddb.totals = newTotals[0]
-      ddb.totalsDiff = newTotals[1]
-      ddb.trend = jsonData.daily
-      ddb.lastUpdated = jsonData.updated[0].lastupdated
-      
-      drawLastUpdated(ddb.lastUpdated)
-      drawKpis(ddb.totals, ddb.totalsDiff)
-      drawPageTitleCount(ddb.totals.confirmed)
-      drawPrefectureTable(ddb.prefectures, ddb.totals)
-      drawTrendChart(ddb.trend)
-      
-      whenMapAndDataReady()
-    })
 
+    function loadDataOnPage() {
+      loadData(function (data) {
+        jsonData = data
+
+        ddb.prefectures = jsonData.prefectures
+        let newTotals = calculateTotals(jsonData.daily)
+        ddb.totals = newTotals[0]
+        ddb.totalsDiff = newTotals[1]
+        ddb.trend = jsonData.daily
+        ddb.lastUpdated = jsonData.updated[0].lastupdated
+
+        drawLastUpdated(ddb.lastUpdated)
+        drawKpis(ddb.totals, ddb.totalsDiff)
+        drawPageTitleCount(ddb.totals.confirmed)
+        drawPrefectureTable(ddb.prefectures, ddb.totals)
+        drawTrendChart(ddb.trend)
+
+        whenMapAndDataReady()
+      })
+    }
+
+    loadDataOnPage()
+
+    // Reload data every five minutes
+    setInterval(function () {
+      unDrawKpis()
+      pageDraws++;
+      jsonData = undefined;
+      initDataTranslate()
+      setTimeout(function () {
+        loadDataOnPage()
+      }, 1000)
+    }, FIVE_MINUTES_IN_MS)
   })()
 </script>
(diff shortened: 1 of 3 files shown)


olingern:
wondering if we can add a formatter like prettier or prettier-standard in a separate PR to reduce the white space diffs and keep spacing consistent?

static/main.js
```diff
@@ -278,21 +278,35 @@ function drawPrefectureTable(prefectures, totals) {
...
   let totalStr = 'Total'
   if(LANG == 'ja'){
     totalStr = '計'
   }
-  
+
   dataTable.innerHTML = dataTable.innerHTML + "<tr class='totals'><td>" + totalStr + "</td><td>" + totals.confirmed + "</td><td>" + totals.recovered + "</td><td>" + totals.deceased + "</td></tr>"
 }
 
+function unDrawKpis(totals, totalsDiff) {
```
reustle: I think I'm missing something here, not sure why we can't just overwrite the numbers with the newly loaded numbers, instead of needing an 'undraw' function?

index.html
```diff
@@ -145,44 +145,60 @@ <h4 data-ja="一次情報源">Primary Data Sources</h4>
...
+
     initDataTranslate()
     drawMap()
-    
+
+    var draws = 0;
     var styleLoaded = false
     var jsonData = undefined
-    
+    var FIVE_MINUTES_IN_MS = 300000
```
reustle: Nitpick: should this be a `const`?

static/main.js
olingern: I do this to show that the data is reloading, so it shows the loading spinners again and then replaces those, i.e. resets to the original page load state before downloading again

index.html
```diff
@@ -145,44 +145,60 @@ <h4 data-ja="一次情報源">Primary Data Sources</h4>
     if(typeof drawMap === 'undefined'){
       return
     }
-    
+
     initDataTranslate()
     drawMap()
-    
+
+    var draws = 0;
```
reustle: Maybe `pageDraws` to be slightly more specific? Especially with a global variable.

index.html
olingern: Yes, I didn't know that level of IE support was wanted, so I was used the existing convention. But, updated to `const`

index.html
olingern: done :+1:

reustle:
> wondering if we can add a formatter like prettier or prettier-standard in a separate PR to reduce the white space diffs and keep spacing consistent?

Sounds like a good idea. I feel we're inevitably sliding towards using webpack (and adding something like React) anyway :)

index.html
reustle: Good call. Looks like at least IE11 supports it (not IE10). Future babel support coming anyway 👌

static/main.js
reustle: Is it simple to leave the page completely intact and just overwrite data / redraw graph / prefectures where necessary? I think it would both simplify the code, and be easier on the user? It seems currently the data disappears for as long as the new network request is being loaded/parsed? (or am I misreading).

Thanks!

olingern: > Is it simple to leave the page completely intact and just overwrite data / redraw graph / prefectures where necessary?

I use it as a [signifier](https://sites.google.com/site/thedesignofeverydaythings/home/signifiers) of sorts. You can test this by uncommenting the `unDrawKpis` on the `index.html` page, if you have it pulled down locally. I had to put a log inside of the `setInterval` to make sure it was working.

Basically, the user doesn't know what's happening unless some visual indication is given. In fact, they'll probably be inclined to refresh the page anyway unless there's some visual countdown to next refresh or a "Last refresh" time

reustle:
Tested locally, looks good! There are few things, like using the 'update' method of the chart instead of redrawing it, but as that will likely be replaced soon it's no big deal.

The only missing piece would be to implement this on `embed.html` as well, as that's what sites like the Japan Times are using to display data on their own sites. Then we should be good to go 🚀 

Thanks for helping out!

olingern:
Done in d6f0106!

reustle:
Hey @olingern 

This seems to be throwing errors for a lot of users, we may need to revert. Can you try to resolve it? Thanks!

image
image

olingern:
@reustle do you know if that originates in `/embed.html` or `/index.html`?

reustle:
My current view shows me its /embed, but the screenshot above doesnt show /embed so I'm guessing its both. Across all browsers

olingern:
Found the problem. Looks like this made through when I was testing

https://github.com/reustle/covid19japan/commit/603acfd419b55e7363b313fe645dc3156d791b74#diff-feb5e67b085cf4eb27b88136983e487bR117

reustle:
Wow you're right. It's reloading every 2 seconds on the news websites. I will fix if no PR in a few min. Yikes

reustle:
I think prefecture borders aren't showing at all anymore? Could you verify? Thanks!

olingern:
I've checked out a much older commit. The map looks the same.

## Previous
image

## Current

image

reustle:
Found the issue. Here's what it used to look like:

image

Issue is here: https://github.com/reustle/covid19japan/pull/64/files#diff-45a12245c90c3f724f1644a02264341eL400-L412

The order of drawing layers is important, and now that drawing the border layer happens before drawing the prefectures instead of after, it gets buried beneath them. I was able to solve locally by instead drawing the borders layer after the prefectures were drawn.

olingern:
Okay, I'm on Arch linux so it could be a version of Firefox / Chrome issue. Currently, firefox looks the same all the way back to ~Feb 20th.

Chrome, however, looks quite broken on my machine. There's a 403 of sorts as far back as 2ccbee2c43f013b0842f93bc55fc537cf2a75be1

2020-03-04_958x865

reustle:
> Chrome, however, looks quite broken on my machine. There's a 403 of sorts as far back as 2ccbee2

I've actually seen a bunch of those errors on Sentry but could never explain them.

PR #68, opened 2020-03-03
Use correct UTF-8 meta tag for HTML5 documents

coliff:
REF: https://github.com/h5bp/html5-boilerplate/blob/master/dist/index.html

@reustle


reustle:
Thanks!

PR #89, opened 2020-03-19
Handle fetch() errors properly

liquidx:
Handle fetch errors with retry with exponential backoff.


reustle:
Thanks! So the HTTP request for the data file fails, fetch throws an error, and we catch it and then show this? Don't think we should auto-retry instead?

liquidx:
Good idea about auto-retry. 

Though there is still the case when there are client errors (read failed)...

liquidx:
Changed the PR to do a simple exponential backoff rather than display an error state.

PR #92, opened 2020-03-19
Copy CNAME into docs/ as part of the build

liquidx:
@reustle 

CNAME keeps on being cleaned out of docs/ on each build because of the clean step.

This PR copies it back into docs/ on each build to save ourselves from having to keep reverting the CNAME delete change.

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,1 +1,0 @@
-covid19japan.com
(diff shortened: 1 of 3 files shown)


reustle:
Great, actually we can delete `/CNAME` because we only need `/docs/CNAME`. It reads from whichever dir is currently live.

Thanks!

Issue #84, opened 2020-03-18
Travel Bans


reustle:
1. Add a list of countries banned from travel into Japan (or need quarantine)
2. Add a list of countries which people coming from Japan are banned from entering

i.e. https://twitter.com/motokorich/status/1240248532377858050
> parts of Spain, Switzerland, and all of Iceland, plus mandating 14-day quarantines for travelers from 38 countries, inducing all Europe, UK, Ireland and Egypt.

**Update**: V1 is now launched on the site, we just need to update the data

---

<img width="696" alt="image" src="https://user-images.githubusercontent.com/304560/77826465-e4511f00-7152-11ea-8544-6c45f8e60ad0.png">

nateberkopec:
Relevant 

https://www.iatatravelcentre.com/international-travel-document-news/1580226297.htm

https://www.japan.travel/en/coronavirus/

ann-kilzer:
These are helpful sources. Wondering what's the best way to summarize this, as it's a long list. Could divide into a few categories, like "No entry", "Must Quarantine", "Visa on arrival suspended." What do you think? I have a WIP branch I can share soon.

reustle:
Another source: https://www.japantimes.co.jp/news/2020/03/19/national/abe-imposes-14-day-quarantine-japan-arrivals-several-countries-europe/#.XnLxsS2B2gQ

Yeah, I think a few simple lists would work, like below. Maybe each country on the list could be a link to the source of that info?

---

### Traveling Into Japan

Below are the travel restrictions that have been put in place regarding traveling to / from Japan. Click on the country link for more details on the restrictions.

**Banned From Entering Japan**: Italy, China, ...

**Existing Visa Required**: Spain, ...

**Other Limitations**: Greece (details), Germany (details), ...

### Traveling Out Of Japan

*Similar lists here, if any*

l15n:
Here are the primary sources

- For denial of landing, via the MoJ http://www.moj.go.jp/hisho/kouhou/20200131comment.html
    - Announcement: http://www.moj.go.jp/nyuukokukanri/kouhou/nyuukokukanri01_00128.html
    - English: http://www.moj.go.jp/content/001316999.pdf
- For 14-day quarantine restrictions, its the MHLW:
    - Announcement: https://www.mhlw.go.jp/stf/seisakunitsuite/bunya/0000121431_00098.html


There's also numbers on

l15n:
I've only noticed two actual categories:
[...]

PR #97, opened 2020-03-22
84 travel restrictions

ann-kilzer:
Taking a first stab at #84 

Since there's so much variance, I decided to make a simple table with the ability to add links for further details. 

- [x] Replace the dummy data with an actual data store. Need some help figuring out where the other table data is populated from
- [x] Make the table look pretty, with alternating shading.

@reustle would love feedback


reustle:
Thanks, it looks good! Yeah, I agree with styling the table. 100% width, and maybe include the same stripes as the prefecture table (every other row). Also guessing you'll make the links in column 3 clickable.

How does the "no entry" table mean compared to the "banned entry" list?

Appreciate it :)

Screenshot attached for others
image

reustle:
Heya, can I help this along in any way? Should we ask others to help pull in the restriction data? Thanks again

ann-kilzer:
I was gonna take another look at the code tonight when I finish work. What would help is to know how to pull in data sources from outside. Tried copying the example from the prefecture table but I feel like I'm missing something.

reustle:
> What would help is to know how to pull in data sources from outside. Tried copying the example from the prefecture table but I feel like I'm missing something.

I think for now it's fine if this is hardcoded, edited via a PR? Pulling data via the spreadsheets is a bit convoluted, and we're mid-write on that piece too (see https://github.com/reustle/covid19japan-data/ )

ann-kilzer:
Okay @reustle fixed the table and adjusted some of the presentation. Ready for another look, but could probably use a hand putting together the rest of the country data for foreign borders.

ann-kilzer:
Ready for another look @reustle !

Could use some help populating the rest of the data, but I got one category completed. 🚢

src/index.js
```diff
@@ -401,6 +574,32 @@ function drawPrefectureTable(prefectures, totals) {
...
+  travelRestrictionsHelper('#foreign-other-restrictions', ddb.travelRestrictions.foreignBorders.other);
+}
+
+function travelRestrictionsHelper(elementId, countries) {
+  let countryList = [];
+  // Iterate through and render country links
+  _.orderBy(countries, 'name', 'desc').map(function(country){
+    let name = (LANG == 'en') ? country.name : country.nameJa;
+
+    countryList.unshift(`${country.emoji}<a href="${country.link}">${name}</a>`);
```
reustle:
```suggestion
    countryList.unshift(`<a href="${country.link}">${country.emoji} ${name}</a>`);
```
Making the flag clickable

src/index.html
```diff
@@ -150,7 +149,25 @@ <h4 data-ja="都道府県ごとのデータ">Prefecture Data</h4>
...
+  <section id="travel-restrictions" class="embed-hide">
+    <h4 data-ja="">Traveling into Japan</h4>
+    <p>Below are the travel restrictions that have been put in place regarding traveling into Japan.
+      Click on the link for more details on the restrictions.</p>
+    <h5>Banned From Entering Japan:</h5><span id="banned-entry"></span>
+    <h5>Existing Visa Required:</h5><span id="visa-required"></span>
+    <h5>14-day Self-Quarantine Required:</h5><span id="self-quarantine"></span>
+    <h5>Other Limitations:</h5><span id="other-restrictions"></span>
+
+    <h4>Traveling out of Japan</h4>
```
reustle: Let's hide this section for right now, until we get some data loaded in

src/index.js
ann-kilzer: Sure, took the space out to prevent odd linebreaks that separate the flag and country name.

reustle:
Oops approved, but we need to build `/docs`

Please run this, then commit, and we should be good: `npm run build-prod`

ann-kilzer:
@reustle ok good to merge!

reustle:
Thanks @ann-kilzer 🎆 🥳

PR #101, opened 2020-03-23
Add link to Ibaraki's infection case listing

l15n:
Adds the following link, which is the actual listing of Ibaraki case details. Despite the URL, it is up-to-date to today (March 23)

https://www.pref.ibaraki.jp/hokenfukushi/yobo/kiki/yobo/kansen/idwr/press/20200318-corona.html

Note: Ibaraki's top level page for coronavirus is:
https://www.pref.ibaraki.jp/hokenfukushi/yobo/kiki/yobo/kansen/idwr/information/other/documents/20200130-corona.html

---

Hi @reustle, adding sources for Ibaraki to the README. Shouldn't conflict with #100.

PR #102, opened 2020-03-23
Add link to Oita Prefecture sources

l15n:
@reustle Adding link to Oita prefecture. Follow-up to #101 
Shouldn't conflict with #100 either


liquidx:
Thanks!

PR #117, opened 2020-03-26
Switch to use data served from data.covid19japan.com 

liquidx:
@reustle Finally, this CL switches the endpoint to data.covid19japan.com. 

Minor format changes:
- Renames some fields to add "*Cumulative" for those that are not just the daily total.
- Simplifies lastUpdated value (not an array any more)
- Renames prefecture to name (because the array is called "prefectures")

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -73,10 +73,10 @@ function loadData(callback) {
     .then(function(data){
       callback(data)
     })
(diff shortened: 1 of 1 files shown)


reustle:
お疲れ様! Thank you!

Issue #122, opened 2020-03-28
Warning On Todays Data


reustle:
On both the trend line and cases bar chart, the values for `today` are quite low until the end of the day.

Let's add some kind of indicator to the user (greyed out color, tooltip, etc) to make it clear that these are not the final totals for the day. Very open for suggestions on the best way to do this.

ann-kilzer:
Lower the saturation of the color? Use an open circle or a dotted line between today's and yesterday's dot?

PR #123, opened 2020-03-28
ADD : Daily provisional value feedback

Crocsx:
@reustle

Hello, 

I kind of discovered the website today and had some free time.

I am PRing for :
https://github.com/reustle/covid19japan/issues/122

It can be improved, but apparently c3 do not support dashed bar chart so there is 3 solutions :

1) line are dashed, bar is colored (what color you would like to use?)
2) line and bar chart are colored (what color you would like to use for each of the last sections ?)
3) hack a bit c3 and make a dashed bar chart (I didn't look into it)

for the PR i made option 1, but it's easy to change to 2.

Also, is `暫定` accurate for `Provisional` in japanese ?

Thank you.

Screenshot_2


reustle:
Great, thank you @Crocsx ! For the 'today' bar, could you please use `rgb(209,214,223)`?

And @Kay-Cube confirmed that the Japanese translation is correct.

Crocsx:
done 
Screenshot_1

reustle:
> In addition to this, I think you should always make the last row be today even when there is no data for today in the exel

@liquidx Do you think this is something we should do on the data compilation side? Better to keep only what exists in the sheet?

ann-kilzer:
What do you think about making the final dot an open circle? ⭕️

reustle:
Great, thank you @Crocsx - could you please also run `npm run build-prod` and commit that? 🙏

Crocsx:
would you like the dotted circle ?

reustle:
@Crocsx Not sure what you mean by dotted circle, but either the current solid circle or a circle with a hole in the middle as Ann mentioned would be good. Just need the docs to be built, primarily, and we can merge it :)

Kay-Cube:
It looks sweat! @Crocsx 
I think the final dot can have a 0.5 opacity (same logic as the bar), to keep visual consistency

Crocsx:
@Kay-Cube  yes looks not bad 

Screenshot_2

reustle:
And it's live! Thanks a bunch :)

PR #124, opened 2020-03-28
Improved localization using json files

### LINKED ISSUE #93: Improved localization using json files
I notice that localization is inline in the code. It might help to switch to a framework that allows json files for easier translation -- i.e., translators can focus on the content and don't need to understand javascript / html.

Suggestions for code frameworks:
- https://github.com/jeresig/i18n-node-2
- https://github.com/AGROSICA/node-localize

Suggestions for translation tools:
- https://lokalise.com/ (use the free trial to kickstart things)

I have friends that can help with Japanese translation.

What do you think, @reustle?

Related: #87

### LINKED ISSUE #133: Japanese language alignment
@reustle 
Align terms of confirmed patients in Japanese language.

New Cases By Day
陽性者数 (日別) --> 日別新着感染者数

Confirmed Case Trajectories by Prefecture (> 50 Cases)
都道府県別の患者数推移 (50例以上) --> 都道府県別の感染者数推移 (50例以上)

mertd:
I implemented the use of json localization files using i18next as the framework. I could add additional translations in a future PR, but I wanted to stay as close as possible to the current version with this PR so I only ported the already translated strings.

Libraries added:
* [i18next](https://github.com/i18next/i18next) is the i18n framework
* [i18next-browser-languagedetector](https://github.com/i18next/i18next-browser-languageDetector) plugs in with the above to automatically detect the browser language or settings from former visits
* [loc-i18next](https://github.com/mthh/loc-i18next) to actually parse the document and populate html elements with `data-i18n` attributes with the translations.

@reustle 
Closes #93 
Closes #133


reustle:
Really helpful, thank you @mertd . If you search for `LANG` in index.js, you'll see we need to use that sometimes to determine which language to render within a component. Sounds like we'll tackle those in a future PR?

Also, there are a lot of formatting changes mixed into this PR, making it quite difficult to review. Do you have any suggestions of how we can clean up this PR, and prevent issues like this in the future? (created #128 )

Cheers

mertd:
@reustle I just cleaned up the PR (reset to the formatting from before while solving merge conflicts) -- sorry for that. I probably let my editor format as a matter of habit. :)

I just tested the language setting some more. `LANG` is kept in sync when language is switched via the UI, but it gets out of sync when the browser is set to `ja`. I will have a look at how to imprive this

reustle:
Thanks!

> reset to the formatting from before while solving merge conflicts

Hm, still seeing the formatting changes in the diff, maybe it wasn't pushed? https://github.com/reustle/covid19japan/pull/124/files#diff-1fdf421c05c1140f6d71444ea2b27638L252-L266

Just let us know when you've handled some of the lang detection and we'll review. Seems you can do a quick check on page load to see whether or not you should 'trigger' the language flip or not to match the browsers default lang. Then, we can go through and remove all of the legacy language toggle stuff in favor of the new method. Happy to hear your thoughts, thanks.

mertd:
I added a `setLang` function and moved most that was in the onClick-Event for the language switch in there, and use it in the i18next initialization as well. The language settings should be synchronized now.

About the formatting: I only reset those that were part of the merge conflicts. I can revert some more or help with #128 so code can be formatted automatically.

reustle:
Do we want to merge the code formatting first, so that the diff on this one is simpler?

mertd:
Makes sense!

mertd:
- [ ] #133

mertd:
Last changes:
* merge and resolve conflicts
* added translation fixed from #133 
* added localization part to Readme.md

reustle:
Thanks @mertd - Is it possible to get rid of these merge conflicts, or will we have to do some PR trickery?

mertd:
@reustle I fixed the merge conflicts. For now, I left translations that were done in the `index.js` untouched, as I have seen their render method can fail in which case the translation might be in a weird mixed state. I would suggest we migrate those seperately.

mertd:
Closes #133

mertd:
One more note: I left the strings within the HTML elements where they were, though they are entirely optional now as they will be overwritten once the localization function gets called. I am not sure if removing them has any impact on SEO though.

reustle:
>  I left translations that were done in the index.js untouched, as I have seen their render method can fail in which case the translation might be in a weird mixed state. I would suggest we migrate those seperately.

Sounds good

> One more note: I left the strings within the HTML elements where they were, though they are entirely optional now as they will be overwritten once the localization function gets called. I am not sure if removing them has any impact on SEO though.

I think it's fine to leave them in for SEO reasons? Google does crawl w/ basic JS, though.

mertd:
@reustle you mentioned possible render issues in #152. Anything in particular I could double check?

mertd:
@reustle merged `master` and fixed conflicts again. I love the netlify deploy previews, by the way!

reustle:
Thank you and sorry for the delay! 

Sometimes I see both language options at once

image


Also sometimes this graph doesn't seem to translate consistently with the rest of the page.

image

Any ideas?

mertd:
Just migrated translations for all manually-localized strings in `index.js` that are not from the `ddb` global variable.

I can reproduce the issue with the graph you are referring to and will investigate. I could not reproduce both language switches being shown at the same time though -- do you have steps to reproduce?

mertd:
@reustle I fixed the chart not being subjected to language change. The graph needed to be redrawn, which is not currently happening in `master` as well.

reustle:
> do you have steps to reproduce?

I don't know, sorry :S I only saw it 2 or 3 times during my times testing it, refresh would fix it. Maybe some kind of cookie that doesn't exist on a fresh reload?

> The graph needed to be redrawn, which is not currently happening in master as well.

Ok let's handle that in a separate PR then.

mertd:
> Ok let's handle that in a separate PR then.
It's fixed in this PR already. :) I simply added the graph's draw function to the functions called when a language change occurs.

liquidx:
LGTM 👍

reustle:
Hmm, translations seem to be totally broken in production, is it just me?

image

mertd:
On https://covid19japan.com translations are broken for me too. Is the deployed page the latest version? I included some changes to the Japanese country names which do not seem to be in production either.

I believe https://covid19japan.netlify.com/ is a newer deployment as language changes work and the country name changes are included too.

Issue #109, opened 2020-03-24
Regional/Prefecture-Level trajectories


l15n:
I've found that FT's cross-country/region trajectory charts are great for understanding the relative status of pandemic spread.

https://www.ft.com/coronavirus-latest

Example via Twitter:
![](https://pbs.twimg.com/media/ET1FmRRWkAI4QKw?format=jpg&name=small)
https://twitter.com/jburnmurdoch/status/1242225650125193219

I think that adding a graph like this would be especially helpful in Japan because of how reporting and disease tactics are not centralized here. There's a big difference between the prefectures that is quite difficult to grasp with just the latest total figres.

What's needed:

- Prefecture-level daily trend data (cases, not deaths right now)
- Offset the above to an arbitrary starting point, e.g days since N cases
- Convert dates to days since N cases
- Chart with log scale on y-axis

This should be achievable with current data, after some degree of massaging. 

@reustle What do you think? This is something I'm interested in building.

liquidx:
That would be awesome. 

We just started a data repo that has exported the data from the spreadsheet into a JSON file:
https://github.com/reustle/covid19japan-data/tree/master/patient_data

I'm in the middle of automating pulling data from the spreadsheet and checking it in, but at the very least, that JSON file should have the data you need.

reustle:
Thoughts of having the page show country-wide by default, with the option to drill down by prefecture?

KPIs and trend charts would show by prefecture, map would stay the same.

Otherwise, we add some kind of sparkline into the existing prefectures table, and make the height fixed and scrollable (so the table isn't as tall as it is now).

l15n:
As a visitor, will the visualization tell me what I want to know? Does it summarize the data in a useful way? I'm not sure, so I'd like to try building a prototype first to make a concrete comparison.

> Thoughts of having the page show country-wide by default, with the option to drill down by prefecture?

Since the page is getting crowded, I do agree that country-wide should be a priority. However, since global dashboards show Japan's top level numbers anyway, getting a handle on local status is likely to be more helpful for residents.

> KPIs and trend charts would show by prefecture, map would stay the same.
> 
> Otherwise, we add some kind of sparkline into the existing prefectures table, and make the height fixed and scrollable (so the table isn't as tall as it is now).

Do you have a good example? The best one I can think of is the country comparison on [Our World in Data's Corona Virus page](https://ourworldindata.org/coronavirus)

[...]

PR #125, opened 2020-03-28
Add Prefecture trajectory visualization

l15n:
Implements https://github.com/reustle/covid19japan/issues/109

Screen Shot 2020-03-28 at 11 22 03 PM

I've set an initial threshold of 50 cases for this chart. This threshold may need to be increased to 100 in coming days.

Hopefully this can give users some context on the relative rate of change of prefectures.

@reustle What do you think? I've tentatively added the chart just above the prefectures table.

In terms of the implementation, there's a fair bit of data manipulation going on. I think it's OK to keep doing this at runtime for now, but if necessary, perhaps the cumulative array can be moved into the data repository.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -132,6 +132,13 @@
     </div>
   </section>
 
+  <section id="prefecture-trajectory-container" class="embed-hide">
+    <h4 data-ja="都道府県別の患者数推移 (50例以上)">Confirmed Case trajectories by prefecture (> 100 cases / logarithmic)</h4>
+    <div id="prefecture-trajectory">
+        <div class="text-center"><div class="lds-dual-ring"></div></div>
+    </div>
+  </section>
+
   <section class="embed-hide">
     <h4 data-ja="都道府県ごとのデータ">Prefecture Data</h4>
     <table id="prefectures-table">
--- a/src/index.html
+++ b/src/index.html
@@ -133,7 +133,7 @@
   </section>
 
   <section id="prefecture-trajectory-container" class="embed-hide">
-    <h4 data-ja="都道府県別の患者数推移 (50例以上)">Confirmed Case trajectories by prefecture (> 100 cases / logarithmic)</h4>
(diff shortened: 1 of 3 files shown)


liquidx:
One thing though, the chart doesn't look logarithmic. as in expected it to compress between 100 to 1000, but it looks linear to me.

reustle:
👏 Great work! :shipit:

l15n:
> One thing though, the chart doesn't look logarithmic. as in expected it to compress between 100 to 1000, but it looks linear to me.

It is intentionally linear right now. Will change it to logarithmic when the numbers get big enough.

l15n:
Screen Shot 2020-03-29 at 1 18 16 AM

I updated the title and styles after rebasing with master.

reustle:
Looks good, ready to merge?

l15n:
> Looks good, ready to merge?

👍 

Though it looks like if I include the production build `docs/index.js` etc, it is going to conflict every time. Should we leave the production build out of pull requests, and instead update separately, either pushing to master or separate pull requests?

l15n:
I removed the conflicting production build so it should be good to merge. Will need a separate commit to build and put it in production

reustle:
> and instead update separately, either pushing to master or separate pull requests?

Hmm, I think everyone else includes the build of /docs within their PRs, as long as its merged with the latest master there shouldn't be any conflicts? Maybe I'm not understanding you, though.

Thanks! I guess we'll have another PR shortly for the /docs bulid?

l15n:
> Hmm, I think everyone else includes the build of /docs within their PRs, as long as its merged with the latest master there shouldn't be any conflicts? Maybe I'm not understanding you, though.

It's a timing issue where, if Pull Request A and Pull Request B both have `/docs`, and Pull Request A gets merged first, the author of Pull Request B will get a conflict until they can pull master and rebuild. This can get cumbersome because the `/docs` will not merge cleanly, so you need to do something like remove the commit where you did a production build, then build again. 

This isn't a difficult procedure, but does mean that the more people are involved, the more often people will have to re-run `npm run build-prod` multiple times 🤷‍♂

PR #126, opened 2020-03-28
Adds a 30-day column sparkline to the prefecture table

liquidx:
This allows showing of the relative trend per-prefecture. Easily compare the prefectures at a glance to see how much each is increasing relative to each other.

This also adds an indicator for how much the count increased per-prefecture, just like the total.

Screen Shot 2020-03-28 at 23 42 24


src/index.js
```diff
@@ -346,17 +345,52 @@ function drawDailyIncreaseChart(sheetTrend) {
   })
 }
 
+function drawSparkLine(elementId, seriesData, maxConfirmedIncrease) {
```
reustle: Thoughts on using something a bit more specific like `drawPrefectureTrendChart`? I think I mistakenly used the word [sparkline](https://en.wikipedia.org/wiki/Sparkline) in our previous chat.

src/index.js
```diff
@@ -377,28 +411,62 @@ function drawPrefectureTable(prefectures, totals) {
...
+    }
     
-    // TODO Make this pretty
-    
-    if(pref.name == 'Unspecified'){
+    if (pref.name == 'Unspecified'){
       // Save the "Unspecified" row for the end of the table
-      unspecifiedRow = "<tr><td><em>" + prefStr + "</em></td><td>" + pref.confirmed + "</td><td>" + (pref.recovered?pref.recovered:'') + "</td><td>" + pref.deaths + "</td></tr>"
-    }else if (pref.name == 'Total'){
+      unspecifiedRow = `<td class="prefecture">${prefStr}</td>` +
```
reustle: I think the new template literals work safely across multiple lines ([source](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)). Is there another reason you wanted to add `+` on each row?

reustle:
Fantastic, thanks! Left a few minor comments in code.

I took this screenshot from your demo, is this still how tooltips are formatted? Maybe we can use:
`3/14: 30 cases` to make the date clearer?
image

src/index.js
liquidx: Good point, I'll just call it trend everywhere.

src/index.js
liquidx: Ohh, I didn't know that. That makes it much nicer ...

liquidx:
I fixed the tooltips to display the date on hover. 

The CPU on usage isn't too bad on mobile, I've tested it and most of the time is spent actually drawing the map  rather than generating the SVG for the lines.

It looks good on mobile now, had to fix some of the size issues.

localhost_4000_(iPhone X)

PR #138, opened 2020-03-29
feat: Show Last Updated As Time Ago

### LINKED ISSUE #132: Show Last Updated As Time Ago
Currently the last updated time looks like:

>  Last Updated: Mar 29 2020, 15:53 JST 

Let's instead show it as:

>  Last updated 28 minutes ago

>  Last updated 3 hours ago

> Last updated 18 hours ago

colinthornton:
Resolves #132

Parses last updated date string and displays a relative time
difference (includes i18n).

Currently expects the date string to
be in the format "MMM DD YYYY, HH:mm z" as in
"Mar 29 2020, 15:53 JST".


liquidx:
Thanks Colin, this looks great!

colinthornton:
@liquidx Thanks! One thing I just noticed though, I made the PR without using moment, but it looks like moment is now being imported in `src/index.js` since commit https://github.com/reustle/covid19japan/commit/7b0ce715a85f526b546f72bd77349da43e4311ce . If moment's going to be used anyway then we can just use their [fromNow](https://momentjs.com/docs/#/displaying/fromnow/) method (comes with localization) and remove the `src/utils/date.js` file I made.

If you'd prefer to use moment then just let me know and I'll make the changes.

liquidx:
Oh, that would be better. Would you be able to change it to use moment.js?  Thanks!

colinthornton:
@liquidx Alright, I switched it over to using moment.js.

src/index.js
```diff
@@ -854,16 +854,33 @@ function drawKpis(totals, totalsDiff) {
...
 
-
+/**
+ * @param {string} lastUpdated - MMM DD YYYY, HH:mm JST (e.g. Mar 29 2020, 15:53 JST)
+ */
 function drawLastUpdated(lastUpdated) {
   // Draw the last updated time
 
+  const display = document.getElementById('last-updated');
+
```
liquidx: Could you check here whether display is undefined before continuing?

colinthornton: That element is in the static markup: https://github.com/reustle/covid19japan/blob/master/src/index.html#L70

Is there some reason it would be removed dynamically?

liquidx: We're seeing errors consistently where elements we thought should be there don't exist (maybe the DOM hadn't finished loading yet).

I'm actively trying to figure out why that is, but meanwhile, we just try to guard things as much as possible. Thanks!

PR #140, opened 2020-03-29
Data Gen Cleanup

reustle:
Now that Alastair has pulled together the great [data repo](https://github.com/reustle/covid19japan-data/), we can remove all data related code from this codebase


reustle:
Wondering if any of the old site visitors (using cached JS) will now stop seeing updates once this is merged 🤔

Issue #158, opened 2020-03-31
Text Alignment Issue


reustle:
There is a small alignment issue happening on the embed page, check the text just below the map. It should be left/aligned with the rest of the content. 

You can see it live here: https://www.japantimes.co.jp/liveblogs/news/coronavirus-outbreak-updates/

---

![image](https://user-images.githubusercontent.com/304560/78000827-d5938380-736f-11ea-933b-2605e11b16e3.png)

PR #160, opened 2020-03-31
[#158] Center embed map footnote

brendanrygus:
Fixes text alignment issue from https://github.com/reustle/covid19japan/issues/158
cc @reustle 

## Proposed Changes
- Convert embed `footnote` to `p` tag so that centered margins will apply to block level element
- Small refactor of `embed-hide` / `embed-show` CSS to reduce specificity collisions
  - ex. Setting `display: block` on the existing footnote `span` would be overwritten by the `body.embed span.embed-show { display: inline; }`
  - This could use a quick IE11 test if you have the capacity, [Can I Use](https://caniuse.com/#search=%3Anot) says that it should work though

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -116,7 +116,7 @@
     <div><span class="four">▉</span> <span data-ja="201+ 件">201+ cases</span></div>
   </section>
   <div id="map-container"></div>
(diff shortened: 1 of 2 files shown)


liquidx:
Looks good, thanks for the fix and making it cleaner!

PR #166, opened 2020-03-31
Only send 5% of errors to Sentry

l15n:
See: https://docs.sentry.io/error-reporting/configuration/?platform=javascript#sample-rate

> Configures the sample rate as a percentage of events to be sent in the range of 0.0 to 1.0. The default is 1.0 which means that 100% of events are sent. If set to 0.1 only 10% of events will be sent. Events are picked randomly.

@reustle If total errors sent to Sentry is a problem, sampling errors is one way to mitigate. Let's try 5%?

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -44,6 +44,7 @@
       Sentry.init({
         dsn: 'https://1c7127f5f5b2432d99a5bf44a3b74d08@sentry.io/2967034',
         debug: false,
+        sampleRate: 0.05,
         integrations: [
           new Sentry.Integrations.CaptureConsole({
             levels: ['error']


liquidx:
Nice find. I think this makes sense. 5% seems reasonable

reustle:
I don't think this has been an issue outside of a few cases where we had something generating many errors. I think a better approach might be to set an alerting level on sentry when we see an increase, and revert it if needed. # of errors into sentry will be sponsored by someone if not sentry themselves, so it's not a large worry there.

I guess I'm just worried about some errors falling through the cracks, or this causing it to take longer for us to notice issues. Not very strongly opinionated, though.

l15n:
> I guess I'm just worried about some errors falling through the cracks, or this causing it to take longer for us to notice issues. 

In my experience using sampling, this is not an issue. When you have a lot of errors, you simply don't notice rare errors, so sampling is actually good enough because you can only really afford to handle the errors with the most volume.. If you can keep errors very low, then there is no need for sampling anyway, and its worth handling all errors.

My main reason for proposing this was because it seemed like the volume of errors was going to be a financial burden. Since that doesn't seem like an issue, there's no urgent need to use this option. Now that we know it is there, we can turn it on when we need it.

PR #167, opened 2020-03-31
Remove unused dependency chart.js

l15n:
We have three charting libraries now: c3.js, chart.js, apexcharts. Might as well eliminate the one we're not using?

@reustle Just some quick cleanup


liquidx:
Thanks!

PR #168, opened 2020-03-31
Remove apexchart dep. Switch trend graph to c3.

liquidx:
This isn't as pretty, but should be safer to use than apexcharts.


liquidx:
Screen Shot 2020-04-01 at 0 09 20

PR #169, opened 2020-04-01
Optimised favicon.png image

tagawa:
Reduced the filesize (21kb -> 12kb) by dropping the colour depth to 255 (+ alpha channel), as it's just a simple image. This might save a little bit of bandwidth.

cc @reustle


reustle:
Thank you!

PR #172, opened 2020-04-02
Fix language selector

liquidx:
Restrict languages to only en and ja.

On some browsers, it's returning en-US and we don't handle that properly.


liquidx:
I think this should fix it. I tried a bunch of different options in the detector (whitelist, etc), but the most reliable way was to only that the first two chars in the language. 

Basically the problem was the detector would return "en-us" and then our code would break because we expect "en"

PR #181, opened 2020-04-03
Start Prefecture Trajectory from 100 cases

l15n:
Now showing from 100 cases instead of 50.

Screen Shot 2020-04-03 at 10 41 16 AM

Also updated the colors, using [a different palette](https://observablehq.com/@d3/working-with-color?collection=@d3/d3-color) and doing some HSL manipulations to grey out some of the prefectures.

FYI @reustle @liquidx


l15n:
Yeah, will keep this implementation and iterate.

Issue #182, opened 2020-04-03
Include daily increase in Confirmed Case Trajectories by Prefecture


colinthornton:
@reustle @l15n 

The Outbreak Spread Trend chart includes the total cases as well as the daily increase, e.g. `1142 (+40)`.
![image](https://user-images.githubusercontent.com/17635021/78316522-d9efb480-759a-11ea-8313-16b3803630e8.png)

It would be nice if the Confirmed Case Trajectories by Prefecture chart displayed similarly by including the daily increase.
![image](https://user-images.githubusercontent.com/17635021/78316698-479be080-759b-11ea-9987-1e79ef9470cd.png)

colinthornton:
I saw lots of news about Tokyo having 97 new cases yesterday, so I wanted to compare our numbers.

l15n:
Good idea! I'll see if I can add it quickly

l15n:
This is now working in production 😄

colinthornton:
@l15n Awesome!

PR #185, opened 2020-04-03
Include daily increase in trajectory tooltip

l15n:
Implements #182 

Similar to the trend chart, shows the daily increase per prefecture in the tooltip.

It's a large-ish change because a lot of the pre-processing needed to be reorganized.

FYI @reustle @colinthornton

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -623,49 +623,46 @@ function drawPrefectureTrajectoryChart(prefectures) {
   const filteredPrefectures = _.filter(prefectures, function (prefecture) {
     return prefecture.confirmed >= minimumConfirmed;
   });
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -674,13 +671,15 @@ function drawPrefectureTrajectoryChart(prefectures) {
...
-    _.values(lastIndex),
+    _.values(trajectories),
     function (a, b) {
-      return Math.max(a, b);
+      return Math.max(a, b.lastIndex);
     },
     0
   );
 
+  console.log(`Max Days: ${maxDays}`);
```
reustle, line 681: Intentionally left?

l15n: Accidental 😓  Will clean it up. Thanks!

PR #186, opened 2020-04-04
Add travel bans from 2020-04-03

mertd:
I also moved the travel band json into its own file as it's gotten quite long

Related: #84 

@reustle


reustle:
You're a star! Thanks so much. Please mark it as 'ready for review' when you'd like it checked. Cheers :)

mertd:
@reustle should be ready :)

src/index.js
```diff
@@ -17,6 +17,9 @@ import locI18next from "loc-i18next";
 import translationEn from "./i18n/en.json";
 import translationJa from "./i18n/ja.json";
 
+// import static data
+import travelRestrictions from "./data/travelRestrictions"; // refer to the keys under "countries" in the i18n files for names
```
ann-kilzer, line 21: Nice 👍

reustle:
Thanks! Do you know if there is a page we can link to instead of a PDF? Also wondering if it makes sense in the near future to link tot he local embassy website for each country (not sure if they're even updating this info yet for some).

mertd:
I guess the PDF is the most official source.

japan.travel also maintains a list though: https://www.japan.travel/en/coronavirus/

PR #193, opened 2020-04-05
テーブルで既存感染者数がある方が便利だと思いますね

TechnicalSuwako:
各県の既存感染者数を見えるのは便利だと思いますので、これを追加してみました。
I think it's convenient to see the number of infected people in each prefecture, so I added this.


reustle:
Thank you! ありがとう！

@liquidx thoughts here? I think we need to think about making the table scroll horizontally without causing the whole page to become wider.

PR #200, opened 2020-04-06
Use 感染者 instead of 患者

l15n:
Fixes an inconsistency in the translations that was initially covered in #133 by @Kay-Cube but was accidentally reverted somewhere along the line.

FYI @reustle


reustle:
Didn't merge because of it only being src/ and not docs/ - I've scheduled to finish Netlify tonight at 22:00 👌

PR #209, opened 2020-04-07
feature / code decomposition

ekreerenko:


ekreerenko:
@reustle, can you please check my PR?

src/data/constants.js
```diff
@@ -0,0 +1,70 @@
+import translationEn from "../i18n/en.json";
+import translationJa from "../i18n/ja.json";
+
+export const TIME_FORMAT = "MMMM d yyyy, HH:mm";
```
liquidx, line 4: Only used in one place, I suggest moving it to where it is actually used.

src/data/constants.js
```diff
@@ -0,0 +1,70 @@
...
+export const COLOR_ACTIVE = "rgb(223,14,31)";
+export const COLOR_CONFIRMED = "rgb(244,67,54)";
+export const COLOR_RECOVERED = "rgb(25,118,210)";
+export const COLOR_DECEASED = "rgb(55,71,79)";
+export const COLOR_TESTED = "rgb(164,173,192)";
+export const COLOR_TESTED_DAILY = "rgb(209,214,223)";
+export const SUPPORTED_LANGS = ["en", "ja"];
```
liquidx, line 11: Only used in one place, also the order of the consts would be better if it was logical or alphabetical.

liquidx:
I have a few comments/requests for the change:

1. Could we have fewer directories? Many of the directories just have only an index.js. This makes editing really hard because my editor will be full of files with the same filename. I suggest not having directories under src/ if we could help it.

2. If we have directories, can all directories start with a lowercase character?

3. Many of the consts in constants.js only get used by a single file, so I'd rather keep them only declared where it's used. Only move shared constants into a common file. 

4. I'd suggest naming constants.js to something like common.js. We don't want to name things like consts, utils because it becomes unclear what the purpose of the content inside. 

4. There is a typo for the directory named OutebreakMap.

src/data/constants.js
```diff
@@ -0,0 +1,70 @@
...
+    en: {
+      translation: translationEn,
+    },
+    ja: {
+      translation: translationJa,
+    },
+  },
+};
+
+export const DDB_COMMON = {
```
liquidx, line 53: Only used in index.js, would be better if this was moved back to where it is being used.

src/pages/OutebreakMap/index.js
```diff
@@ -0,0 +1,16 @@
+import { MAP_CONFIG } from "../../data/constants";
+
+export const drawMap = (mapboxgl, map) => {
```
liquidx, line 3: This could be better if it was merged with drawMapPrefectures.js and just contained all the things that makes the map work. Splitting the two parts to initialize the map and draw the prefectures seem like it's a little too much separation.

src/data/helper.js
```diff
@@ -0,0 +1,53 @@
+export const calculateTotals = (daily) => {
```
liquidx, line 1: How about name this file something different than helper.js? This only contains a calculation of totals, how about just called it calculateTotals.js ?

src/pages/DailyIncrease/index.js
```diff
@@ -0,0 +1,125 @@
+import * as c3 from "c3";
```
liquidx, line 1: How about naming this DailyIncreaseChart? This makes it a noun rather than a verb. 

It's what the method was named and it makes it consistent with naming the files for what they are creating (like OutbreakMap and TrendChart)

src/pages/SpreadTrend/index.js
```diff
@@ -0,0 +1,146 @@
+import * as c3 from "c3";
```
liquidx, line 1: Suggest naming this SpreadTrendChart (consistent with making it clear what it's trying to draw)

src/pages/TravelRestrictions/index.js
```diff
@@ -0,0 +1,38 @@
+import i18next from "i18next";
+import _ from "lodash";
+
+const travelRestrictionsHelper = (elementId, countries) => {
+  const countryList = [];
+  // Iterate through and render country links
+  _.orderBy(countries, "name", "desc").map((country) => {
```
wapgear, line 7: If using only `orderBy` in this file it'll be better to have 
```
import { orderBy }  from "lodash";
```
instead of importing the whole lodash

src/data/constants.js
wapgear: My opinion is to store constants in a separate file, so even if it was used only once it'll be better to store it with other constants, otherwise it'll be double work every time when this value will be needed again

src/pages/PrefectureTable/index.js
```diff
@@ -0,0 +1,198 @@
+import * as c3 from "c3";
+import _ from "lodash";
```
wapgear, line 2: Instead of importing all `lodash` methods it'll be better to import only what is actually needed

wapgear:
@liquidx 
> Could we have fewer directories? Many of the directories just have only an index.js. This makes editing really hard because my editor will be full of files with the same filename. I suggest not having directories under src/ if we could help it.


Sorry, but having directories in src is totally fine, btw if you’ll open few index.js files in most of editors it’ll collapse index.js part in name and will show different part - directory name, so it should be possible to enable such feature in settings
example: 
image

What about content in `index.js` files and `lowerCase` directory names - do this repo has a convention about code style? I think it's a good case to add it and describe these cases in this

src/data/constants.js
liquidx: If a constant is only used in one place, what would be the advantage of pulling it into a separate file? 

Keeping it in the same file can make the code more readable.

liquidx:
> @liquidx
> 
> > Could we have fewer directories? Many of the directories just have only an index.js. This makes editing really hard because my editor will be full of files with the same filename. I suggest not having directories under src/ if we could help it.
> 
> Sorry, but having directories in src is totally fine, btw if you’ll open few index.js files in most of editors it’ll collapse index.js part in name and will show different part - directory name, so it should be possible to enable such feature in settings
> example:
> image

I prefer less hierarchy, less navigation and less redundancy. Maybe I'm not seeing the advantage of the increase in heirarchy. What might be a good thing that comes from the increase in hierarchy?


> 
> What about content in `index.js` files and `lowerCase` directory names - do this repo has a convention about code style? I think it's a good case to add it and describe these cases in this
> 
> cc @reustle

l15n:
> I prefer less hierarchy, less navigation and less redundancy

:+1: on this.

At least here, there is only one page so far so I think that `pages` is a poor name. Right now all the charts map closer to components.

src/data/constants.js
reustle: But this is more of a configuration option than simply a string that is used, no?

python273:
Agree with having less folders and index.js files. Seems redundant in case there's only one source file in a folder and index.js like
```
├── Kpi
│   ├── drawKpis.js
│   └── index.js
```
could be reduced to `drawKpis.js` basically

Regarding constants, I think if it's expected to use it as configuration, it's fine to move to one file

src/data/constants.js
liquidx: TIME_FORMAT is only used in one function

reustle: That's fine, as @python273 also pointed out, its more of a configuration option, so moving it makes sense. Someone might fork this and 'time format' is a common thing that you may want to configure.

reustle:
🎉 Deployed, thanks everyone. Will keep an eye on Sentry for the next hour~

Issue #207, opened 2020-04-07
Force Page Reload


reustle:
Force the page to reload after 1 hour. We currently auto refresh the data every few minutes.

PR #213, opened 2020-04-10
Automatic reload

l15n:
Resolves: #207 

A very basic automatic reload. Defaults to 3 hours right now, but can be changed.

Also accepts `refreshInHours` parameter to manually set a different reload time in hours. Must be >= 1.

@reustle This is a bit of a band-aid approach, but best to get the ball rolling first

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -216,6 +216,16 @@ const loadDataOnPage = () => {
   });
 };
 
+const startReloadTimer = () => {
+  const params = new URLSearchParams(location.search);
+  let reloadInterval = parseInt(params.get("refreshInHours"));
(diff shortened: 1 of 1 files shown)


reustle:
Looks good! Not sure if we really need the URL param parsing feature, though? Seems it's just for testing, and that can be done by modifying the value?

l15n:
> Not sure if we really need the URL param parsing feature, though? Seems it's just for testing, and that can be done by modifying the value?

Yeah, its probably not entirely necessary. I'll pull it out to keep things simple.

l15n:
Thanks for following up on this one @reustle

PR #216, opened 2020-04-10
Convert flags to SVG with twemoji

### LINKED ISSUE #208: Convert Flag Emojis To Unicode
Reports are it doesn't work in Windows. It seems we may need to provide our own icons... I don't unicode will be enough. Thoughts?

https://github.com/lipis/flag-icon-css

mertd:
Added twemoji (as recommended by @l15n) and a parser for flag images that will replace emoji with svg images on Windows only.

Notes and open questions:
- I did not yet build as I have high hopes for #201 🚀 
- Should we only use twemoji flags for Windows or make the app behave the same everywhere?

@reustle 
Closes #208


reustle:
> I did not yet build as I have high hopes for #201 🚀

Sorry, this is already done, but the PR is just to remove the leftovers :) Production is using Netlify, but some cached DNS was still using github pages. Will merge soon.

> Should we only use twemoji flags for Windows or make the app behave the same everywhere?

My vote would be to have the same everywhere.

Thank you!

mertd:
@reustle nice! I removed the os check

reustle:
LGTM

image

reustle:
The only remaining comment is, sometimes the flag is on a different line than the country name. When it was simply emojis, this didn't happen. This is actually why there is currently no space between the flag and name. Do you see some simple way that we can make sure the flag and name are always on the same line? If not, no worries. Maybe we can at least add the space between the flag and the name?

🙏 thanks

mertd:
grafik
Like this? :)

PR #217, opened 2020-04-10
Make prefectures table scrollable

### LINKED ISSUE #203: Scrollable Prefectures Table
Right now on mobile, the prefectures table is too wide, causing the entire page to scroll left/right. Let's make it so only the prefectures table can scroll horizontally.

mertd:
@reustle 
Closes #203


reustle:
Great, thank you! Do you think we should set a min-width on these columns so that the headers don't mash together?

image

mertd:
grafik
This how it looks like with `min-width: 7em;` on `th`

reustle:
Great, thanks. Could you please resolve the new conflict and then we'll get it merged. 🙇

Issue #215, opened 2020-04-10
Updated Header


reustle:
via @Kay-Cube 

New:
![image](https://user-images.githubusercontent.com/304560/78968776-602e6c80-7b40-11ea-8a4c-5124cd30e919.png)

Current:
![image](https://user-images.githubusercontent.com/304560/79034243-76dfcc80-7bef-11ea-9162-b2cbc7a02893.png)

bmaniteja:
How can I Help ?

reustle:
@bmaniteja Thanks for offering to help :) Are you able to make these changes on the UI?

bmaniteja:
Please review https://github.com/reustle/covid19japan/pull/224

PR #218, opened 2020-04-10
Spanish translation

kant:


mertd:
There will be two additional changes needed to make the app usable in a new language:
- Import your translation in `src/data/constants.js` and add to resources in the `LANG_CONFIG` export
- Add to language switcher (@reustle maybe #215 should automatically display all available locales?). Until then, Spanish would be rendered only if the user's browser is set up to display Spanish, or if the page is accessed with `lng=es` as a query parameter

reustle:
Thank you for translating everything!

Yeah I'm sure about adding many languages to the switcher (see #215 for UI) but if it automatically sets itself to Spanish if the users browser is set to Spanish, then they can manually click English / Japanese if they want. Does that sound good?

kant:
@reustle I trust in the whole project, so, go ahead with _your_ plan. Whatever it is your choice, it will be `OK` for my persona.

mertd:
Since #227 is merged, you only need to add "es" to the languages array in `src/i18n/index.js` so that the app actually reads your translation file.

mertd:
#259 is merged and will require you to add "es" to the `LANGUAGES` array as well as a flag emoji to the `FLAGS` array in `src/i18n/index.js`

mertd:
Should I make the necessary changes? I could fork from @kant's branch so their contribution is counted.

reustle:
Sounds good @mertd

PR #225, opened 2020-04-11
Fixes overflowing Prefectures table making only table scroll

bmaniteja:
@reustle 
Fixes https://github.com/reustle/covid19japan/issues/203

output

### DIFF
--- a/src/index.scss
+++ b/src/index.scss
@@ -38,7 +38,7 @@ section, header, footer {
   padding: 30px 10px;
   max-width: 650px;
   margin: 0px auto;
-  overflow-x: scroll;
+  overflow-x: auto;
 }
 .center-offset {
   max-width: 650px;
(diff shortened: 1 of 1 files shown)


reustle:
Thank you! I think this is already fixed in https://github.com/reustle/covid19japan/pull/217 ?

But your implementation seems simpler. Thoughts @mertd ?

Thanks both

mertd:
This would be nicer because it doesn't need an additional wrapper element, but it also adds scrollbars for me to a lot of other elements:
grafik

reustle:
Yeah it seems this shouldn't be on `section, header, footer {` but instead that table specifically

bmaniteja:
> This would be nicer because it doesn't need an additional wrapper element, but it also adds scrollbars for me to a lot of other elements:
> grafik

oh wow i didn't realise this let me look into this.

bmaniteja:
Fixed it by making auto so that we don't see default scroll

reustle:
I've just merged the other table width fix PR so I think we can safely close this one. Thanks again!

PR #226, opened 2020-04-11
Added Twemoji To Language Selector

bmaniteja:
@reustle crated this PR incase we decide to go with twemoji.

Screen Shot 2020-04-11 at 5 49 48 PM

https://github.com/reustle/covid19japan/issues/208

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -1,3 +1,0 @@
-Oops! You are currently viewing the old version of <a href="https://covid19japan.com">covid19japan.com</a>.
-<br/><br/>
-Please refresh this page to try again.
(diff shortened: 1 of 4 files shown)


reustle:
Thank you! Just merged the earlier Twemoji PR so that likely cleans up what you need to implement?

mertd:
Since we already have twemoji as a dependency in the package.json, I suggest removing the script tag loading it and using the managed dependency instead by moving the twemoji.parse call into an appropriate JS file. :)

reustle:
Hey, merged the new header PR, I think this PR can be simplified now? Thanks!

reustle:
Hey @mertd would you please be able to suggest some changes on this one, or launch a new one as this one has been silent for 10~ days?

Thanks to you both!

mertd:
@reustle If this is inactive, I could open another PR. I would include some changes to populate the language selection from the central array so that adding new languages doesn't require changes in the selector as well. Thoughts?

reustle:
Sounds good, thanks. We can also incorporate that into #218 then too.

PR #229, opened 2020-04-12
Limit Chart Time Period To 60 Days

reustle:
image

### DIFF
--- a/src/components/SpreadTrendChart/SpreadTrend.js
+++ b/src/components/SpreadTrendChart/SpreadTrend.js
@@ -7,6 +7,7 @@ import {
   COLOR_CONFIRMED,
   COLOR_RECOVERED,
   COLOR_DECEASED,
+  CHART_TIME_PERIOD,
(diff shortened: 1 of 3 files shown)


src/components/SpreadTrendChart/SpreadTrend.js
```diff
@@ -20,7 +21,11 @@ const drawTrendChart = (sheetTrend, trendChart) => {
     Tested: ["Tested"],
   };
 
-  for (let i = 0; i < sheetTrend.length; i++) {
+  for (
+    let i = sheetTrend.length - CHART_TIME_PERIOD;
```
liquidx, line 25: Minor: This is a little bit weird format for a for-loop, but I guess the sheetTrend.length - CHART_TIME_PERIOD is too long?

reustle: I wrote it as one line, but it looks like the auto-cleanup tool made this change :)

Issue #141, opened 2020-03-30
Show number of cases and projections with logarithmic y-axis.


aywander:
@reustle , plotting `log(number_of_cases)` makes it easier to see whether cases are increasing exponentially or not. Doubling times can also be over-plotted as straight lines for comparisons, like [this example in ourworldindata.org](https://ourworldindata.org/coronavirus#are-deaths-increasing-at-different-rates-in-different-countries).

reustle:
I've seen some trackers where you can switch between our current view and logarithmic. Thoughts @liquidx @l15n ?

l15n:
Using the logarithmic scale has been on my mind for a while as well. Up until very recently, the cases in Japan were growing linearly, so it made sense to stay linear. While linear scale, its still easy to identify when a curve is shifting to be exponential. 

Here's what it looks like when logarithmic.

## Trend Chart for Japan

![Screen Shot 2020-03-30 at 10 26 13](https://user-images.githubusercontent.com/3714/77867218-51f36d00-7271-11ea-815c-0eb64b42d407.png)

Starting from 1.

## Trajectory Chart by prefecture

![Screen Shot 2020-03-30 at 10 28 27](https://user-images.githubusercontent.com/3714/77867209-4c962280-7271-11ea-87c7-5153ba3ac440.png)

Starting from 100

Note, c3.js is a little broken here because `type: "log"` is still buggy and doesn't calculate the data-level label positions properly.

----

I havent' figured out how to add guidelines to c3.js yet.

aywander:
@l15n, you're right, both linear and logarithmic are useful. @reustle, A toggle would be indeed be great.

aywander:
> Note, c3.js is a little broken here because type: "log" is still buggy and doesn't calculate the data-level label positions properly.

Here also, perhaps the easiest is just to create a parallel data set of the logarithm of the data and plot that on a linear y-axis.

reustle:
Thanks @aywander - For now, we'll leave it as is. There are likely better ways to display the data, but I think high level trends are probably where we may end up. Appreciate taking the time to leave feedback.

PR #232, opened 2020-04-14
Quick tiny modification to show spread graphs in decimal or log scales

prajdabre:
@reustle I have currently put a button to toggle between decimal and log scales. Would this be what you are looking for?

If this is ok then I can do this for all other graphs. Just let me know which ones.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -1,3 +1,0 @@
-Oops! You are currently viewing the old version of <a href="https://covid19japan.com">covid19japan.com</a>.
-<br/><br/>
-Please refresh this page to try again.
--- a/src/components/PrefectureTable/PrefectureTable.js
+++ b/src/components/PrefectureTable/PrefectureTable.js
@@ -108,6 +108,7 @@ const drawPrefectureTable = (prefectures, totals, prefectureTrendCharts) => {
   });
 
   // Iterate through and render table rows
+  let pref_ctr=0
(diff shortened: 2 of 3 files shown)


liquidx:
Thanks for the PR, we really appreciate it.

I took a look at the log graph, while maybe interesting mathematically, it seems like a straight line and not that easy to understand for the average person. We want to keep the graphs simple and understandable.

Unfortunately, I think we will not merge this feature in right now. But again, thank you so much for the PR.

prajdabre:
Hello,

That is perfectly ok. I made this PR because I noticed that someone had raised an issue (#141 ).
Im happy to contribute in any way :)

vlumi:
> We want to keep the graphs simple and understandable.

As a user I would really appreciate the option to see the graphs in logarithmic scale, as the current "Confirmed Case Trajectories by Prefecture" is largely unreadable because the majority of the prefectures are flat on the X axis, due to Tokyo overwhelming the numbers. With logarithmic scale the early increases could be compared between prefectures, to see whether they follow a common curve or not.

Linear scale should probably be the default, as it is more intuitive for seeing the current state.

reustle:
Hello @prajdabre ! Thanks so much for this PR. Unfortunately it didn't get merged, and it is now very far from master. I will close this for now, but we still greatly appreciate the contributions. Thank you.

PR #234, opened 2020-04-15
List relative patient numbers in prefecture table

norbusan:
This change adds a column in the prefecture table listing the numbers of cases per 100000 inhabitants. The population data is based on the 2015 census as shown on the Wikipedia page (see constants.js).

@reustle


norbusan:
The column headers probably need some "reformulation", not sure how to squeeze this information into a short string. Maybe "10ppm" (10 times parts per million)?

colinthornton:
Also it's not clear if the term "cases" is equivalent to the "confirmed" or "active" numbers to the left.

norbusan:
> Also it's not clear if the term "cases" is equivalent to the "confirmed" or "active" numbers to the left.

Indeed, I used confirmed, but rethinking the "meaning" of the number "active" would probably better, since this is the current density of infected person.

norbusan:
Maybe this should be done?
```
diff --git a/src/components/PrefectureTable/PrefectureTable.js b/src/components/PrefectureTable/PrefectureTable.js
index fb95ec8..1c01b56 100644
--- a/src/components/PrefectureTable/PrefectureTable.js
+++ b/src/components/PrefectureTable/PrefectureTable.js
@@ -178,7 +178,7 @@ const drawPrefectureTable = (prefectures, totals, prefectureTrendCharts) => {
         <td class="count">${pref.deceased ? pref.deceased : ""}</td>
         <td class="count">${pref.active || ""}</td>
         <td class="count">${(
-          pref.confirmed / PREFECTURE_POPULATION_IN_CHUMAN[pref.name]
+          (pref.active || 0) / PREFECTURE_POPULATION_IN_CHUMAN[pref.name]
         ).toFixed(2)}</td>
       `;
       prefectureTrendCharts = drawPrefectureTrend(
@@ -201,7 +201,7 @@ const drawPrefectureTable = (prefectures, totals, prefectureTrendCharts) => {
           totals.confirmed - totals.recovered - totals.deceased
         }</td>
         <td class="count">${(
-          totals.confirmed / PREFECTURE_POPULATION_IN_CHUMAN["Japan"]
+         (totals.confirmed - totals.recovered - totals.deceased) / PREFECTURE_POPULATION_IN_CHUMAN["Japan"]
         ).toFixed(2)}</td>
         </tr>`;
```

norbusan:
And the table head could be renamed "density" / "密度" ?

norbusan:
I did the changes discussed above:
- change column title to Density 密度
- compute the value via active number instead of confirmed
- give a one line explanation below the table how density is computed

liquidx:
Thanks for the PR! This is pretty interesting.

However, there is already a lot of numbers in that table and I'm reluctant to add more to the table unless it's very useful. The per-capita information could be a little misleading as the rate of infection generally has not much to do with population sizes and not that easy to understand.

norbusan:
Hi @liquidx 

> However, there is already a lot of numbers in that table and I'm reluctant to add more to the table unless it's very useful. 

Ok, I see your point. But then, dropping the "confirmed" in favor of the "active" would be probably advised, too, since (1) the number is a combination of the rest, (2) not as interesting, since it contains recovered cases.

> The per-capita information could be a little misleading as the rate of infection generally has not much to do with population sizes and not that easy to understand.

Are you sure? As a mathematician I would disagree. Of course, population density in the area is also of importance, but in general I assume that the density of infections has a direct relation to the probability to get in contact with an infected person, and thus with the probability to get infected, too.

All the best

liquidx:
> Hi @liquidx
> 
> > However, there is already a lot of numbers in that table and I'm reluctant to add more to the table unless it's very useful.
> 
> Ok, I see your point. But then, dropping the "confirmed" in favor of the "active" would be probably advised, too, since (1) the number is a combination of the rest, (2) not as interesting, since it contains recovered cases.

Agreed, there already is too much data we're showing (it used to only be confirmed and deaths), because we used to not even have recovered cases count per-prefecture until maybe two weeks ago. 

> 
> > The per-capita information could be a little misleading as the rate of infection generally has not much to do with population sizes and not that easy to understand.
> 
> Are you sure? As a mathematician I would disagree. Of course, population density in the area is also of importance, but in general I assume that the density of infections has a direct relation to the probability to get in contact with an infected person, and thus with the probability to get infected, too.

:) I'm not a mathematician, just a normal person, so maybe I'm completely wrong. Though, I look at numbers all day long in my real job. My intuition on it is that size of population has something to do with the maximum confirmed cases (e.g. where is the peak?), rather than the rate. The rate, as you say, seems to be primarily about density, or maybe more specifically the rate of contact amongst people (like a packed train or department store). 

The other trouble we have is that prefectures are not uniformally dense/sparse. Most prefectures have big cities and small cities.

> 
> All the best

norbusan:
Ok, fine with me. Thanks for your consideration. I am closing this PR.

Issue #237, opened 2020-04-15
CPU and Memory Usage


reustle:
We're having quite a but of memory usage (leaks) and cpu usage over time. Would love some help debugging and optimizing, as unfortunately I'm quite swamped with work right now. We can discuss anything related in this card. Thank you!

a-ogilvie:
@reustle Can you share the results of any investigation you've done so far?

reustle:
Thanks for reaching out, I haven't dug too deep but here's what I know:

- Sometimes we see 'memory usage limit reached' on sentry errors list for a select few users, probably who've left the page open for a long time. We've added an auto page refresh every few hours for the time being. Likely some sort of memory leak

- The initial page load has become significantly slower. Map, tons of charts / graphs, etc is likely the case of this. It would be nice to ensure that there aren't any inefficient calculations happening at least. We then need to revisit the idea of removing many of the charts or simplifying them. If it's slow on my laptop, its definitely slow on phones.

Feel free to chime in @liquidx

liquidx:
I'm guessing there might be something we're doing inefficiently with rendering the charts.

I suspect processing a 100~ rows of data isn't really the bottleneck on the browser. But we do get out of memory errors, which may mean that there's a memory/reference leak somewhere we're not catching when the page is loaded for a long period of time.

l15n:
Here are some memory usage snapshots on the same page over time.

Initial load:
<img width="1254" alt="Screen Shot 2020-04-17 at 7 35 06 PM" src="https://user-images.githubusercontent.com/3714/79560526-b708e880-80e2-11ea-862e-edeae4f1e705.png">

About 20 minutes later:
<img width="1256" alt="Screen Shot 2020-04-17 at 7 35 27 PM" src="https://user-images.githubusercontent.com/3714/79560536-bcfec980-80e2-11ea-8648-f41c11a1eaa2.png">

The clear big change here is an increase in memory usage from [ArrayBuffer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer) objects. This strongly suggests that we're somehow failing get rid of old charts after data is updated.

jonathanpperry:
Hi Leonard, out of curiosity what program did you use to capture these memory usage snapshots?

l15n:
> Hi Leonard, out of curiosity what program did you use to capture these memory usage snapshots?

[...]

PR #240, opened 2020-04-18
Performance improvements: Use DCL, separate macrotasks

l15n:
@reustle @liquidx Here's a first attempt at performance improvement for https://github.com/reustle/covid19japan/issues/237  through changes in timing and taking advantage of how the JS event loop runs in a browser.

## Before

Screen Shot 2020-04-18 at 11 01 49 PM

This is a capture of a run through the profiler on `master` using a local environment. My observations were:

* Most of the work doesn't start until the `load` event. Lots of idle time prior to that.
* There's a huge macrotask (labeled Task) after `load` where all the action happens. This blocks all rendering because the event loop is stuck on that task and can't do anything else. The browser won't render until the task completes

I also noticed that `drawPrefectureTable` was called twice because of i18n.

## Approach

So, my basic strategy was:
1. Trigger on `DOMContentLoaded` rather than `Load`. `Load` is actually late because it waits for stuff like images as well. For JS, you only need the DOM to be ready. This saves us some time
2. Split up the big macrotask so that other events can be processed, and hopefully the browser can render parts of the page rather than waiting until the whole thing is finished.
  * This is achieved by using `setTimeout` so that each logical render is a separate macro task in the event loop task queue. 
3. Reduce the coupling between data load and render as much as possible.
4. Try to avoid drawing when there is no data, or no update.

## Result

Screen Shot 2020-04-18 at 10 53 51 PM

(If you look closely, you may notice that the total length of this profile snapshot is actually longer. Ignore that: it's just a measurement artifact)

* Data load happens earlier now, at DCL
* No longer have the unnecessary double render of the trajectory chart
* The slowest charts below the load will render later, so that first view (the numbers) should load a little faster now.
* When data loads, it *should* avoid unnecessary redraws. In the long run, this should hopefully mitigate the memory issues somewhat.

## Further thoughts

Unfortunately, this change doesn't actually reduce the total CPU/memory burden too much. My understanding of the profiler results tell me that:
* c3.js/d3.js is very intensive. If we really want to have interactive charts, they may be worth putting on a separate page. It's where all the CPU time is going.
* data pre-processing probably won't help too much. It will be good for clarity of the code, which will again make it easier to change charting libraries.
* For the prefecture table, I suspect there's some layout thrashing going on. One approach could be to actually build the inner html string for the entire table and insert in one go. Then, draw the trend charts. This will reduce the number of times we write to the DOM and speed it up a bit.
 
## Does it feel faster?

Please try out the staging version and see if it feels better. Rather than my profiler screenshots, I think that actually trying it out should be the ultimate way to judge whether this is an actual performance improvement.

### DIFF
--- a/src/components/PrefectureTable/PrefectureTable.js
+++ b/src/components/PrefectureTable/PrefectureTable.js
@@ -63,6 +63,11 @@ const drawPrefectureTrend = (
   return prefectureTrendCharts;
 };
 
+// Run CPU intensive processing in a separate macrotask
+const enqueueMacrotask = (callback, delay = 0) => {
+  setTimeout(callback, delay);
+};
+
 const drawPrefectureTable = (prefectures, totals, prefectureTrendCharts) => {
   // Draw the Cases By Prefecture table
   const dataTable = document.querySelector("#prefectures-table");
@@ -131,14 +136,14 @@ const drawPrefectureTable = (prefectures, totals, prefectureTrendCharts) => {
         <td class="count">${pref.deceased ? pref.deceased : ""}</td>
         <td class="count">${pref.active || ""}</td>
         </tr>`;
-      setTimeout(() => {
+      enqueueMacrotask(() => {
         prefectureTrendCharts = drawPrefectureTrend(
           `#Unspecified-trend`,
           pref.dailyConfirmedCount,
           maxConfirmedIncrease,
           prefectureTrendCharts
         );
-      }, 0);
+      });
     } else if (pref.name == "Port Quarantine" || pref.name == "Port of Entry") {
       // Override Port Quartantine name as "Port of Entry". The name in the spreadsheet is
       //  confusing.
@@ -154,14 +159,14 @@ const drawPrefectureTable = (prefectures, totals, prefectureTrendCharts) => {
         <td class="count">${pref.deceased ? pref.deceased : ""}</td>
         <td class="count">${pref.active || ""}</td>
         </tr>`;
-      setTimeout(() => {
+      enqueueMacrotask(() => {
         prefectureTrendCharts = drawPrefectureTrend(
           `#PortOfEntry-trend`,
           pref.dailyConfirmedCount,
           maxConfirmedIncrease,
           prefectureTrendCharts
         );
-      }, 0);
+      });
     } else if (pref.name == "Total") {
       // Skip
     } else {
@@ -177,14 +182,14 @@ const drawPrefectureTable = (prefectures, totals, prefectureTrendCharts) => {
         <td class="count">${pref.deceased ? pref.deceased : ""}</td>
         <td class="count">${pref.active || ""}</td>
       `;
-      setTimeout(() => {
+      enqueueMacrotask(() => {
         prefectureTrendCharts = drawPrefectureTrend(
           `#${pref.name}-trend`,
           pref.dailyConfirmedCount,
           maxConfirmedIncrease,
           prefectureTrendCharts
         );
-      }, 0);
+      });
     }
     return true;
   });
--- a/src/components/PrefectureTable/PrefectureTable.js
+++ b/src/components/PrefectureTable/PrefectureTable.js
@@ -131,12 +131,14 @@ const drawPrefectureTable = (prefectures, totals, prefectureTrendCharts) => {
         <td class="count">${pref.deceased ? pref.deceased : ""}</td>
         <td class="count">${pref.active || ""}</td>
         </tr>`;
-      prefectureTrendCharts = drawPrefectureTrend(
-        `#Unspecified-trend`,
-        pref.dailyConfirmedCount,
-        maxConfirmedIncrease,
-        prefectureTrendCharts
-      );
+      setTimeout(() => {
+        prefectureTrendCharts = drawPrefectureTrend(
+          `#Unspecified-trend`,
+          pref.dailyConfirmedCount,
+          maxConfirmedIncrease,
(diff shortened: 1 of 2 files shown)


src/components/PrefectureTable/PrefectureTable.js
```diff
@@ -131,12 +131,14 @@ const drawPrefectureTable = (prefectures, totals, prefectureTrendCharts) => {
         <td class="count">${pref.deceased ? pref.deceased : ""}</td>
         <td class="count">${pref.active || ""}</td>
         </tr>`;
-      prefectureTrendCharts = drawPrefectureTrend(
-        `#Unspecified-trend`,
-        pref.dailyConfirmedCount,
-        maxConfirmedIncrease,
-        prefectureTrendCharts
-      );
+      setTimeout(() => {
```
l15n, line 134: FYI: Using `setTimeout` like this is basically a hack to create a new macrotask into the queue.

src/index.js
```diff
@@ -136,11 +146,13 @@ const setLang = (lng) => {
       if (document.getElementById("travel-restrictions")) {
         drawTravelRestrictions(ddb);
       }
-      prefectureTrajectoryChart = drawPrefectureTrajectoryChart(
-        ddb.prefectures,
-        prefectureTrajectoryChart,
-        LANG
-      );
+      if (ddb.isLoaded()) {
```
l15n, line 149: This is a guard to prevent the trajectory chart from drawing when we first load the page and `i18next` gets run for the first time. This generally runs before data is loaded, so it gets skipped.

src/index.js
```diff
@@ -157,8 +169,9 @@ const initDataTranslate = () => {
     });
 
   // Language selector event handler
-  if (document.querySelectorAll("[data-lang-picker]")) {
-    document.querySelectorAll("[data-lang-picker]").forEach((pick) => {
+  const langPickers = document.querySelectorAll("[data-lang-picker]");
```
l15n, line 172: tiny optimization: don't use querySelectorAll twice.

src/index.js
```diff
@@ -183,36 +196,18 @@ const loadDataOnPage = () => {
...
+    if (!ddb.isLoaded() || jsonData.updated > ddb.lastUpdated) {
+      ddb.previouslyUpdated = ddb.lastUpdated;
+      ddb.lastUpdated = jsonData.updated;
+      ddb.prefectures = jsonData.prefectures;
+      let newTotals = calculateTotals(jsonData.daily);
+      ddb.totals = newTotals[0];
+      ddb.totalsDiff = newTotals[1];
+      ddb.trend = jsonData.daily;
+
+      let event = new Event("covid19japan-redraw");
```
l15n, line 208: I'm using an event to split data loading from rendering. This means that the rendering will happen in a separate macro task and will get scheduled later.

src/index.js
```diff
@@ -236,19 +230,63 @@ window.onload = () => {
...
-  // Reload data every five minutes
-  const FIVE_MINUTES_IN_MS = 300000;
-  const recursiveDataLoad = () => {
-    pageDraws++;
-    loadDataOnPage();
-    setTimeout(recursiveDataLoad, FIVE_MINUTES_IN_MS);
-  };
+// Call only if ddb is updated.
+// Uses a setTimeout to queue a new macrotask
+const callIfUpdated = (callback, delay = 0) => {
```
l15n, line 246: another application of the setTimeout hack explained earlier. Here it's combined with a guard against the ddb being updated, based on the timestamp. It's not perfect, since it isn't actually comparing the actual data. We could get fancy and actually track whether each data source has changed, depending on what is needed for the chart.

src/index.js
```diff
@@ -236,19 +230,63 @@ window.onload = () => {
...
+        prefectureTrendCharts
+      );
+    }, 32);
+    callIfUpdated(() => drawTravelRestrictions(ddb));
+  }
+
+  callIfUpdated(() => whenMapAndDataReady(ddb, map));
+});
 
+document.addEventListener("DOMContentLoaded", () => {
```
l15n, line 286: Start at DCL instead of Load. Also have tried to make this as simple as possible because its the real entrypoint.

l15n:
Doing some testing, I noticed that the map isn't drawing correctly: missing the prefecture names and surrounding geography. I try to figure out where I messed up.

liquidx:
> Doing some testing, I noticed that the map isn't drawing correctly: missing the prefecture names and surrounding geography. I try to figure out where I messed up (unless.. maybe the map only draws correctly in production?)

Did you check whether mapbox is 403'ing you? If I try from localhost it sometimes doesn't load for me either. I think it's mapbox doing blocking based on the hostname for the API key.

src/components/PrefectureTable/PrefectureTable.js
liquidx: Optional suggestion: This can be a little more readable if you wrap the setTimeout() call in another function called something like enqueueMacroTask(). That would mean the call would look like :

```
const enqueueMacroTask = (fn) => {
  setTimeout(fn, 0);
}

... 
enqueueMacroTask(() => { 
   some work here
})
```

That way when you read the code you don't have to guess why there's a setTimeout here.

liquidx:
I can't quite tell if there's a perf diff compared to the current prod site. But it does seem more efficient with the startup.

l15n:
> Did you check whether mapbox is 403'ing you? If I try from localhost it sometimes doesn't load for me either. I think it's mapbox doing blocking based on the hostname for the API key.

Yeah, I think that's it.

PR #243, opened 2020-04-22
Added percentages to the KPI display for quick analysis

sonicviz:
@reustle Added %'s to KPI as I wanted a quick numeric analysis and got tired of doing it by hand

### DIFF
--- a/src/components/Kpi/DrawKpis.js
+++ b/src/components/Kpi/DrawKpis.js
@@ -5,10 +5,56 @@ const setKpiDiff = (key, value) => {
   ).innerHTML = `( ${diffDir}${value} )`;
 };
 
+const setKpiPerc = (key, value) => {
+  document.querySelector(`#kpi-${key} .perc`).innerHTML = `( ${value}% )`;
+};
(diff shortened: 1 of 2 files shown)


reustle:
Thanks for taking the time to do this, @sonicviz ! It seems your editor may have done some auto formatting, making it a bit difficult to see the actual code diff. Could clean that up so we don't accidentally miss any significant changes / bugs? There is also a small merge conflict currently.

🙏 Thanks again

Here's a screenshot for others checking out this pr:

image

sonicviz:
hmm, not sure where those came from. The only changes were to the   <section id="kpi"> where I added         <div class="perc">&nbsp;</div> to the end of each metric.

liquidx:
Yesterday we pushed a change that calculated percentages for some of the key metrics. 

This should resolve the feature of showing percentages and context for some of the numbers.

sonicviz:
Why did you remove the critical amount? That's a key figure among the lot of them as it's directly related to ICU bed capacity.

PR #246, opened 2020-04-22
initEvent requires all 3 arguments

l15n:

https://sentry.io/organizations/reustle-co/issues/1626417201/?project=2967034&query=is%3Aunresolved

Oops, turns out that [initEvent](https://developer.mozilla.org/en-US/docs/Web/API/Event/initEvent) had required arguments 🤦 

(The Sentry issue highlights the line for `createEvent`, but that only has one argument. It's `initEvent` that needs the args)

Should switch to a proper polyfill. Will do that later once I find one.

cc: @liquidx

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -211,7 +211,7 @@ const loadDataOnPage = () => {
       } else {
         // Fall back to deprecated methods for IE11 etc
         event = document.createEvent("Event");
-        event.initEvent("covid19japan-redraw");
+        event.initEvent("covid19japan-redraw", true, false);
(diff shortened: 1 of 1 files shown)


l15n:
Actually, I think I will switch to use CustomEvent and its polyfill instead. Couldn't find an obvious polyfill for the plain Event, though.

https://www.npmjs.com/package/custom-event-polyfill

Source: https://github.com/kumarharsh/custom-event-polyfill/blob/master/polyfill.js

PR #251, opened 2020-04-25
Add 7-day average trend line to new cases.

liquidx:
Adding a 7-day average line to smooth out daily fluctuations.


liquidx:
Screen Shot 2020-04-25 at 9 50 54

PR #257, opened 2020-04-25
Fix non-closed span tag in index.html

python273:


reustle:
Good catch, thanks @python273

PR #258, opened 2020-04-26
Embed JavaScript Widget

### LINKED ISSUE #35: Embed Javascript Widget
We currently have https://covid19japan.com/embed which allows news sites to embed the map directly into their page. Since iframes are difficult, the content is either cut off or too wide. There are [many resources](https://stackoverflow.com/questions/17838607/making-an-iframe-responsive) on how to create responsive iframes. I feel the widget should work as follows

Provide some kind of javascript snippet such as 
``` html
<script src="https://covid19japan.com/embed.js"></script>
```

Which contains roughly the following code
``` javascript
document.write('<div id="covid19japan-embed"><iframe src="https://covid19japan.com/embed"/></div>')
document.getElementById('covid19japan-embed').style = // TODO
```

Basically we create a wrapper element, load the iframe within it, and set the wrapper style to handle the responsiveness properly.

mertd:
Use is documented in the Readme.

You can use the `embed-test.html` for easier testing and review when running the app locally with ´npm run start-webpack´. However, before merging

- [ ] `embed-test.html` should be removed
- [ ] `src/embed.js` lines 13-15 should be shortened to `const src = "https://covid19japan.com/embed";`

@reustle 
Closes #35


webpack.common.js
```diff
@@ -1,78 +1,72 @@
...
-};
\ No newline at end of file
+            loader: "sass-loader",
+            options: { implementation: require("node-sass") },
+          },
+        ],
+      },
+    ],
+  },
+};
```
mertd: Most of these changes are from running prettier; the only semantic change is adding the `src/embed.js` to the build.

reustle:
We should have the Japan Times migrate to this method, yeah?

mertd:
@liquidx Thanks, I removed the test files/lines after your review. :)

dlinx:
I think another quick solution could be returning building KPI image on the server and returning it.

mertd:
@reustle Since the js embed also just uses the `/embed` route, the current iFrames pointing at it should keep working just fine. For automatic resizing, they will need to switch over or process the `setHeight` message though.

@dlinx as far as I am aware, there currently is no server that we have control over since the app is a SPA hosted on Netlify.

PR #259, opened 2020-04-26
Generate language selection from available languages

mertd:
I went with the least invasive approach:
- left hard coded list in the HTML that will be overwritten
- having the languages in one object would be more elegant than one array for languages and emoji each, but it would require additional changes in a lot of other places or a compatibility fallback too

@reustle 
Closes #226


liquidx:
Thanks, this looks great!

PR #262, opened 2020-04-27
Better context for case numbers in map

ChrisLoer:
Hi @reustle ! I live in Japan and really appreciate the work you're doing to help make the covid outbreak a little more intelligible. I wanted to suggest an improvement to the Outbreak Map, and I figured a PR would be the best way to do it. My basic concern is that color coding prefectures based on total number of cases is not as informative as it could be, because (1) if you're looking at the intensity of an outbreak, you really care about per-capita numbers, not absolute numbers, and (2) big prefectures (e.g. Hokkaido) grab all the attention and it's hard to see the real action in say Tokyo, and (3) it's hard to pick up large dynamic ranges on a color scale.

So here's what I tried:

1) Adjust the color-coding to be based on cases-per-million
2) Add extrusions to each prefecture to show a linear indicator of case count

Here's what it looks like:

Screenshot 2020-04-27 15 22 07

I'm not a designer, and I'm sure the thickness/scaling/color-choice/wording can be improved, but I thought I'd submit it like this so if you like the idea it would be easy for you to adapt and pull in.

What I really found myself wanting to do is modify the data-generating tool to export per-prefecture daily counts, and then run the map as a time based animation (I think rolling average of new cases might be more interesting than cumulative counts, but I guess there are a lot of things you could play with). On the other hand, I'm locked upstairs with little kids downstairs and pretending to do _real_ work, so I decided I should scale back my ambitions for now.


ChrisLoer:
Should I create a linked issue? Sorry, not sure how best to engage or who to tag (@liquidx ?)

liquidx:
Thank you for sharing that new visualization. 

Unfortunately, at the moment, it's unlikely we're going to merge this PR because we're already having performance issues with the site and considering simplifying or even making the map static to help the site load faster on low network speed and slower devices.

ChrisLoer:
Ah OK. There may be some low-hanging fruit on running it client side (shouldn't be _too_ expensive), but I agree this probably doesn't need to be interactive so the static maps API could work well. Feel free to tag me if you have questions about using the Mapbox stuff (or would like me to try working on some part of it). I was on the mapbox GL core team for a few years so I'm pretty familiar with at least the rendering side of it.

liquidx:
Thanks for the offer, Chris! We may take you up on the offer to figure out how to optimize our Mapbox GL JS embed. It seems we just used whatever is out of the box so we didn't take time to optimize it.

PR #273, opened 2020-04-29
Popup usability

ekerik220:
@reustle 

Made the popups a little more compact, and prevented mouse getting inside a popup and causing it to "stick" and feel clunky.

Before:
before

After:
after


reustle:
Wonderful, thank you!

liquidx:
Thanks, that definitely improves it.

Issue #270, opened 2020-04-28
[Lighthouse] Reduce JS size


liquidx:
Lots of time is spent blocking on JS parsing and execution. 

<img width="803" alt="Screen Shot 2020-04-29 at 0 41 33" src="https://user-images.githubusercontent.com/72062/80507797-32f01400-89b2-11ea-852d-423aeff0e9a5.png">

Using webpack bundle analyser:
$ npx webpack --config webpack.prod.js --profile --json > stats.json
$ npx webpack-bundle-analyzer stats.json

<img width="1143" alt="Screen Shot 2020-04-29 at 0 42 10" src="https://user-images.githubusercontent.com/72062/80507899-531fd300-89b2-11ea-86fd-7dbbb12bffb8.png">

It looks like our d3, c3 and core-js are the main culprits. Let's see if we can selectively import any c3, d3, or figure out other ways to reduce the size of the JS.

reustle:
Going to close this for now, as you tagged this issue in quite a few optimization PRs

PR #274, opened 2020-04-29
Optimize date-fns and core-js imports

liquidx:
Using a combination of babel plugins and manually going through the imports and testing on IE11, this PR reduces the JS binary size (prod) from 793K to 713K. 

Also removes some slightly buggy manual polyfill that seemed to exist in the codebase with the proper core-js supported one.


liquidx:
Related to #270

PR #278, opened 2020-04-30
Hide travel restrictions sections when empty

ekerik220:

Hide individual travel restriction categories if they're empty. Also hide the entire travel restrictions section if all categories are empty.


liquidx:
Thank you!

PR #287, opened 2020-05-01
Replace c3 trend charts with server rendered ones.

liquidx:


liquidx:
Seems like this change will speed up the page quite a bit, and actually makes the SVGs render nicer than what c3 does at runtime.

Before:
Screen Shot 2020-05-01 at 19 23 42

After:
Screen Shot 2020-05-01 at 19 23 35

PR #288, opened 2020-05-01
Spanish translation (originally by @kant)

mertd:
Spanish translation by @kant with the necessary additional changes so the app knows to load and display it.

grafik

Side note: I noticed I had to add spanish support to `src/components/Header/DrawLastUpdated.js`. It would lower the hurdle for less technical people to help with adding new languages if this could be automated.

Closes #218 
@reustle


liquidx:
This is great. I think it's missing the change to index.html to add the ES link (otherwise while it's loading the list of languages are different.

mertd:
> This is great. I think it's missing the change to index.html to add the ES link (otherwise while it's loading the list of languages are different.

Is the hardcoded list in the index.html not kind of deprecated since we render dynamically based on the available languages now? I can add it for the time being or remove the hardcoded list entirely.

liquidx:
It flashes up on load. I think we should keep it in the static HTML for consistency so that when it first loads (before JS executes), the links are still there so the page doesn't move around i18n kicks in.

mertd:
I made the change, though I think it would be better to have a placeholder text ("Loading languages..." or something similar) to reduce redundancy.

liquidx:
Thanks! Merging the PR. i18next (or somewhere else in the code) does seem to translate that header. But that looks ok for now.

PR #305, opened 2020-05-05
Move national charts up above the map.

liquidx:
Update titles of the chart now that they are side by side (otherwise it's not clear that the first one is a total aggregate number, the other is a daily chart)

localhost_4000_ (11)

localhost_4000_ (8)

localhost_4000_ (7)


liquidx:
Tagging you to see if you have any feedback on this one.

PR #308, opened 2020-05-05
Add colors for axis label on trend line.

liquidx:


liquidx:
In preparation for something like this.

Screen Shot 2020-05-05 at 15 59 45

PR #310, opened 2020-05-05
Resize and layout prefecture tables, maps and charts.

liquidx:
This PR finishes making the full page responsive layout by merging the map and table into a single row on wide layouts.

Collapses the table to only 10 prefectures by default and allows expanding if you need to see more detail.

Various cleanups of the CSS to use rem for all font sizes.


liquidx:
deploy-preview-310--covid19japan netlify app_ (1)
deploy-preview-310--covid19japan netlify app_

PR #312, opened 2020-05-06
Move date-fns locales to i18n exports

### LINKED ISSUE #289: Making it easier to add new languages
When making changes to #218 in #288 to add Spanish as a new language, I noticed that it is not enough to follow the steps for adding new languages as I documented in the Readme, at least for full support of the "Last updated" info. Instead, some changes that go slightly beyond adding a new value to an array have to be made in `src/components/Header/DrawLastUpdated.js`.

Additionally, the new language needs to be added to `src/index.html` if the list should look as similar as possible before and after the automatic language selector generation kicks in.

I suggest we either document the additional changes necessary, or try to lower the hurdle for less technical people who want to help with translations and remove the need to edit two more files.

@reustle

colinthornton:
@liquidx @mertd 
Closes #289 

This is what I came up with for simplifying the process of adding new translations. Unfortunately I had to add a LOCALES export to the `src/i18n/index.js` file, meaning there's another small step for translation contributors to handle, but all the other related things were able to be automated outside that file.

### DIFF
--- a/src/components/Header/DrawLastUpdated.js
+++ b/src/components/Header/DrawLastUpdated.js
@@ -1,13 +1,13 @@
 import i18next from "i18next";
 import { formatDistanceToNow, parse, parseISO } from "date-fns";
-import { enUS, ja, es } from "date-fns/locale";
 
+import { LANGUAGES, LOCALES } from "../../i18n";
 import { TIME_FORMAT } from "../../data/constants";
 
 /**
  * @param {string} lastUpdatedString - MMM DD YYYY, HH:mm JST (e.g. Mar 29 2020, 15:53 JST)
  */
-const drawLastUpdated = (lastUpdatedString, lang) => {
+const drawLastUpdated = (lastUpdatedString, currentLanguage) => {
   // Draw the last updated time
   // If this is called before data is loaded, lastUpdated can be null.
   if (!lastUpdatedString) {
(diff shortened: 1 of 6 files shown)


src/components/Header/DrawLastUpdated.js
```diff
@@ -36,36 +36,36 @@ const drawLastUpdated = (lastUpdatedString, lang) => {
...
+  display.textContent = i18next.getResource(
+    currentLanguage,
     "translation",
-    "last-updated-time",
-    relativeTime["ja"]
+    "last-updated-time"
   );
+};
+
+function addRelativeTimeLocalization(lastUpdated, language) {
```
liquidx, line 53: Minor request. We're using the following function declaration syntax in most places:

const fn = (args) => {
}

Would you mind switching it from function name(args) { } to that format?

liquidx:
That definitely simplifies things, thanks for making the PR

src/components/Header/DrawLastUpdated.js
colinthornton: Done here: https://github.com/reustle/covid19japan/pull/312/commits/98e3175c02046b90f822527b23a61b93d2843e93

`const` declarations aren't hoisted so I had to move those to the top of the file.

liquidx:
Thanks!

PR #318, opened 2020-05-09
Add Indonesian Localization Create id.json

jiprastyo:
@reustle just added Indonesian localization .json file.


liquidx:
Thanks for adding Indonesian translations!

PR #326, opened 2020-05-10
Added Polish translation.

czater:
Hi @reustle!
I added a Polish translation, please take a look 😄


src/i18n/index.js
```diff
@@ -1,10 +1,10 @@
 import { enUS, ja, es, de, ptBR, fr, id } from "date-fns/locale";
```
liquidx, line 1: I think this is missing the pl import here.

liquidx:
Thank you for the PR. I tried to get this to work but it looks like there's a missing import in the PR. 

If you can update the PR and test it with the "deploy preview" link in the "Checks" list to verify, I can merge this in.

src/i18n/index.js
czater: That's right! I fixed it.

czater:
I just updated it! Now it works fine 💯

liquidx:
Thank you so much!

PR #332, opened 2020-05-13
Added Finnish translation.

vlumi:
Hi @reustle

I added Finnish translation for the tracker, please check.


liquidx:
Thanks for the translations and PR. Looks good, I'm going to merge this in.

PR #336, opened 2020-05-15
Fix prefecture trajectory chart to use translated prefecture names

### LINKED ISSUE #335: Lacking prefecture name translation on "Confirmed Case Trajectories by Prefecture"
Hi @reustle 

I noticed that the "Confirmed Case Trajectories by Prefecture" does not have the translated names within the chart itself, only in the tool tip and the legend -- easy to see when switching the language to Japanese:

before

vlumi:
Hi @reustle 

This fixes #335 by using the translated the prefecture name internally, too, when collecting per-prefecture trajectory data. There does not appear to be any use for the original, English name of each prefecture in the chart, as any translated name can be used as the index just as well.

Before:
before

After:
after


liquidx:
Thank you so much for the fix!

PR #339, opened 2020-05-15
Translation of the website to arabic

mftatfcyclone:
I translated all words on the website from English to Arabic. Furthermore, I would like to contribute more to translation or programming in order to publish the website with Arabic language.


src/i18n/ar.json
```diff
@@ -0,0 +1,195 @@
...
+    "peru": "بيرو",
+    "philippines": "الفلبين",
+    "poland": "بولندا",
+    "portugal": "البرتغال",
+    "qatar": "قطر",
+    "romania": "رومانيا",
+    "russia": "روسيا",
+    "saintkittsandnevis": "سانت كيتس ونيفيس",
+    "sanmarino": "سان مارينو",
+    "saudiarabia": "Saudi Arabia",
```
liquidx, line 176: Is there a reason why this one is not translated?

liquidx:
Would it be possible to add this change into change #338 so that we can test the whole change in one PR?

src/i18n/ar.json
mftatfcyclone: I am sorry for that, I have corrected it

mftatfcyclone:
Is there any additional task required for this language? I think all files are ready.

mftatfcyclone:
Is there any additional task required for this language? I think all files are ready.

vlumi:
> Is there any additional task required for this language? I think all files are ready.

The file `i18n/index.js` would need to be updated to refer to the added language, too, matching how the other languages are configured. Also, country translations apart from "westerdam" should be removed, as they come now from another module, with Arabic support.

While with this everything is translated, the flow of everything on the page is still left-to-right -- is that not a problem?

Issue #345, opened 2020-05-18
Use external module for country name translations


vlumi:
Hi @reustle

Considering the trouble of merging all country name translations when new countries are added, and since it's essentially a solved problem, wouldn't it be better to use an existing module to retrieve the translations for the user-selected language?

vlumi:
Added #346 with a proposed implementation.

PR #346, opened 2020-05-18
Use i18n-iso-countries for country name translations.

vlumi:
Hi @reustle

Using an existing country name localization module (i18n-iso-countries) to implement #345.

All countries in `travelRestrictions.json` have been added the two-letter ISO 3166-1 country code, which should make using any other module for the translation easy, too.

The implementation falls back to the i18next translation for cases where a matching country name was not found, necessary for irregular case e.g. "Westerdam".


liquidx:
Thanks for this great PR. 

@mertd - what do you think about this? This might be an improved way to pull in internationalized country names.

mertd:
@liquidx The current country name keys are quite arbitrary (e.g. "ivorycoast" vs "cotedivoire"), so I think this is a great improvement. I wouldn't mind going one step further and replacing the country name keys with ISO country codes altogether.

vlumi:
Thanks for the input! I have further cleaned up and refactored the localization, changing the precedence order for clean override and fallback:

1. i18next
2. i18n-iso-countries
3. The existing `country.name`

The original `country.name` property is not used for anything other than fallback, and could just as well be removed if so desired. For now, I left it in the JSON as just the country code and flag can be a bit cryptic for some countries, making the list harder to maintain if there is no more human-readable property.

Also, I fixed the sorting of the list to be done by the localized name, though it does not support proper i18n sorting rules -- just a standard case-sensitive string sort.

liquidx:
Thank you so much for updating the PR. I'm going to go ahead and merge this in, looks like a good simplification in the translations.

PR #343, opened 2020-05-17
Bangla translator added

sunkuet02:
I have added Bangla translator to this because many Bangla native speaker lives in Japan and follows this site for the updated news. 

@reustle would you please review the changes? Thank you very much for the wonderful website.

PR #342, opened 2020-05-16
Add new travel bans

mertd:
Closes #340
The countries added to travel bans are from the source in the issue above

@reustle

PR #348, opened 2020-05-23
Fix country list in Bangla

vlumi:
Hi @reustle

I have updated the Bangla translations from #343 to match the new structure introduced in #346.

Since the `i18n-iso-countries` module does not support Bangla, I have kept the country translations, and also added the new ones introduced in #342 (with the names copy-pasted from the Bangla wiki, as I don't understand the language).

I have also fixed a crash caused by trying to load an unsupported language in `i18n-iso-countries`, catching the error gracefully and falling back to English country names.


liquidx:
Thank you for finding and fixing the issue!

Looks good, I'll merge this.

PR #349, opened 2020-05-23
Add Hindi Translation

midroid:
Hi @reustle

Creating a PR for Hindi Language Support. Felt it might be helpful for people who need information in Hindi/Nepali/Sanskrit/Marathi or other Indic languages because it shares the similar script for writing.

File Added: src/i18n/hi.json

File Updated: src/i18n/index.js

Also, Please find the attached screenshots for the UI/Visual changes.

I want to point out a potential bug because of using countries name translations module.
No translation file for "hi" and "bn" yet. It was not there initially, so possibly I will raise a  similar PR there also, but it will take some time.

Referenced Module:  https://github.com/michaelwittig/node-i18n-iso-countries/tree/master/langs

Screen Shot 2020-05-23 at 13 59 18
Screen Shot 2020-05-23 at 13 59 58
Screen Shot 2020-05-23 at 14 00 39

PR #350, opened 2020-05-23
Fix country list in Hindi

vlumi:
Hi @reustle

I have updated the Hindi translations from #349 to match the new structure introduced in #346.

Since the i18n-iso-countries module does not support Hindi, I have kept the country translations, and also added the new ones introduced in #342 (with the names copy-pasted from the Hindi wiki, as I don't understand the language).


liquidx:
Thanks for the fix, merged the PR

Issue #351, opened 2020-05-25
Missing translations


vlumi:
Hi @reustle 

I noticed that some languages are falling out of sync from English/Japanese, with missing entries on the localization files. Maybe a new ticket could be opened when new strings need to be localized?

* "languages": missing from cs, de, es, fi, fr, id, it, pl, pt
* de: missing "trajectory-description"
* es: missing "increment-today", "increment-yesterday", "daily", "7-day-average", "daily-trend"
  * moved or deleted from the master: "port-of-entry", "unspecified", "daily-trend"

ghujka:
Hello! I translated into Turkish but have no coding experience. I'm attaching the file if anyone want to implement to the site. 
[tr.txt](https://github.com/reustle/covid19japan/files/5577746/tr.txt)

vlumi:
> Hello! I translated into Turkish but have no coding experience. I'm attaching the file if anyone want to implement to the site.
> [tr.txt](https://github.com/reustle/covid19japan/files/5577746/tr.txt)

Thank you for the contribution! I have created a pull request #466 for the change, to be checked and merged by the maintainers. Also, your translation did not specify the Turkish word for "Turkish" itself (for the language selection), so I picked "Türkçe" from Wikipedia.

A few new translations have been added (and are still missing from pretty much all languages), and would be nice to have for Turkish, also:

"time-period"
"all-time"
"three-months"
"active-cases-description"

ghujka:
time-period: zaman aralığı
all-time: tüm zamanlar
three-months: üç aylık
active-cases-description: aktif vaka tanımı

Thank you very much!

vlumi:
> time-period: zaman aralığı
> all-time: tüm zamanlar
> three-months: üç aylık
> active-cases-description: aktif vaka tanımı
> 
> Thank you very much!

Thanks @ghujka! Sorry, the texts I put are just the keys, the actual English words/sentences are like this:

    "time-period": "Time period",
    "all-time": "All time",
    "three-months": "3 months",
    "active-cases-description": "Total patients that are still currently testing positive for COVID-19.  (Total confirmed cases minus total recovered cases).",

Maybe the first three are as-is, but the first is a bit more complex -- it's a description for the active cases graph.

sp-devi:
HI @reustle @liquidx,
I wanted to contribute and add translations for Filipino language. 

Note; Not sure if this is the correct issue for this comment. If not, apologies.

vlumi:
[...]

PR #352, opened 2020-05-25
Add missing Finnish translations

vlumi:
Hi @reustle 

Added the missing Finnish translation in #351.


liquidx:
Thanks for the updates

PR #355, opened 2020-05-27
Updated i18n-iso-countries to the latest version

vlumi:
Hi @reustle

The latest version of `i18n-iso-countries` (5.5.0) supports Bengali and Hindi, so the local country name translations can be removed and no longer require additions when new countries are added to the travel restrictions list.


liquidx:
Thanks for the pull request! This is great that we can reuse the translations.

PR #357, opened 2020-06-01
Fix current travel bans

vlumi:
Hi @reustle

I noticed that Armenia was missing from the travel bans list.

Also, since there is an official MOJ page for all travel bans currently in effect, I thought it would be better to use that as the link for all of the countries, instead of news articles that just report changes to the bans. The majority of the links were pointing to the MOJ site already.


liquidx:
Thanks for updating the list of links. It is very much appreciated!

PR #360, opened 2020-06-05
I18n number formatting

vlumi:
Hi @reustle 

Now that some of the numbers are getting bigger, I thought that adding thousand separators would improve legibility. Since such formatting is largely locale-dependent, it's better not to implement it from scratch.

I have added formatting of almost all numbers on the page using the standard Intl.NumberFormat, which is supported by pretty much all current browsers:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat

A couple minor things are still missing the formatting:
* The legend on the Outbreak Map
  * Would probably be better to make the boundaries dynamic first -- just constants or automatically derived from data?
* KPI trend chart
  * Burned into the .svg, so needs changes on the other end

I have tested that the formatting works for all of the current languages, on the latest basic browsers for macOS (Safari, Chrome, Firefox), with no issues.

The formatting of Bengali and Hindi use local script for numbers on Safari, while on Chrome and Firefox Hindi uses Arabic numerals. I know neither language so I don't know if there is any problem with this, or which would be the most natural formatting -- using "en_IN" for either would use the same formatting, but with Arabic numerals, if that would be preferred.

Screen Shot 2020-06-05 at 11 29 29

Screen Shot 2020-06-05 at 11 30 26

Screen Shot 2020-06-05 at 11 29 41

Screen Shot 2020-06-05 at 11 29 56


vlumi:
I have included changes to the logic for coloring the map, creating the legend and color boundaries from the same constant.

The numbers on the legend are now i18n-formatted. Also, the localization for the legend items is now more dynamic, with the `from` and `to` values (and just `from` for the last one) parameterized into the localization string (`cases-range`, `cases-last`), to allow language-specific formatting. All language files have been updated to use the new keys, with the current value matching the previous formatting; I changed Japanese to use "〜" as the range character.

Screen Shot 2020-06-09 at 17 57 54

Screen Shot 2020-06-09 at 17 58 02

reustle:
Thanks! Looking forward to getting this reviewed and merged.

vlumi:
Since Iwate is still at 0, I thought it odd that its color isn't on the legend at all -- so I added the "no cases" label there, too, with a color at least close to the actual one.

Screen Shot 2020-06-30 at 10 33 12

There are two somewhat unrelated change sets in this PR now (number formatting and dynamic map legend), but since the latter uses the former I haven't separated them -- let me know if you'd prefer to have them separately and I'll untangle them.

vlumi:
I have this branch currently deployed at https://covid19japan.misaki.fi/ if anyone is interested to have a look.

src/index.js
```diff
@@ -179,6 +181,16 @@ const setLang = (lng) => {
           prefectureTrajectoryChart,
           LANG
         );
+        PrefectureTable.drawAllPrefectureTable(
```
liquidx, line 184: We previously didn't need to redraw the table here. What is the reason we need to draw the tables here now?

liquidx: I see now why you need to do this, to format the numbers to the right language on language change. Looks good.

liquidx:
Merging this PR. Looks great, thanks for putting the effort in to making this work.

src/index.js
vlumi: Yes, the number format may change when the language changes, and (unlike with translations) there are no original numbers nor hooks kept around, so they just have to be re-rendered.

PR #364, opened 2020-07-02
Added Ukrainian localization

nagornyi:
All strings are translated into Ukrainian. I'm a native Ukrainian speaker.
@reustle


liquidx:
Thanks for contributing the Ukrainian translations!

Issue #365, opened 2020-07-03
Allow changing time period of main charts


vlumi:
Hi @reustle 

With the second wave coming, it would be nice to see the full time scale and compare with the first wave -- especially for "Total Cases" and "Daily Confirmed Cases". However, the charts are currently truncated to the last 60 days only, so only the tale of the first wave is visible.

Only showing the recent trend is probably a good default behavior (to see current status at a glance), but having the option to show since the beginning (or another longer/shorter period) would be a nice option to have.

liquidx:
I added a quick selector at the start of the chart section to draw the graph over 6 months and 3 months so you can see the trends. #454

PR #366, opened 2020-07-03
Make the time period for the DailyIncrease and SpreadTrend dynamic

vlumi:
Hi @reustle 

As a first step towards #365, I did some small changes in the backend to make the two charts' time period more dynamic, controlled by the variable `CHART_TIME_PERIOD` (defaulting to the old default renamed to `DEFAULT_CHART_TIME_PERIOD`).

Positive values of `CHART_TIME_PERIOD` behave like before, while negative or zero instead use the full time scale.

No attempt to creating a UI component to control its value (or trigger re-rendering the charts) was made, but just with this it's easier to hack the value with a debugger.

Screen Shot 2020-07-03 at 13 11 13

### DIFF
--- a/src/components/DailyIncreaseChart/DailyIncrease.js
+++ b/src/components/DailyIncreaseChart/DailyIncrease.js
@@ -7,11 +7,16 @@ import { LOCALES } from "../../i18n";
 import {
   COLOR_TESTED,
   COLOR_TESTED_DAILY,
-  CHART_TIME_PERIOD,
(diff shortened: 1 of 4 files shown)


liquidx:
Good idea, I think this is a good improvement. I'm supportive of merging this change at least to give us the ability to either tweak the time period, or possibly make it dynamic. 

Right now, for simplicity, we made it a fixed period, but we could try making it a little smarter or allow people to switch between the two.

liquidx:
Thanks for the PR!

PR #372, opened 2020-07-04
Add two new charts to visualize active cases, recoveries and deaths in better detail.

liquidx:
Rather than just displaying a single chart, attempt to put charts in to context by overlaying active/recovery/deaths against each other to observe any trends.

Still work in progress.


liquidx:
Screen Shot 2020-07-04 at 23 35 33

liquidx:
Obsoleted by #454

Issue #359, opened 2020-06-05
Spanish translation file error


Luistecnologico:
The file that containing the resources in Spanish is missing the line corresponding to "top-prefectures".
the following line should be added `"top-prefectures": "Top 10 Prefecturas",`
@reustle

vlumi:
There are other missing Spanish translations in #351 :
* In language selector: "Languages"
* Column titles in Top 10 Prefectures: "Today", "Yesterday"
* Legend in Nuevos casos diarios: "Daily", "7-day average"

I can create a pull request for these if you (or someone else) can translate them.

reustle:
I was notified of these changes as well by LMendo

> Please change “Activo” into “Activos” it’s weird all the other headers are in plural except that one

> Also, “Prefecturas” should not be capitalized unless at the beginning of a sentence. Therefore “Top 10 Prefecturas” should be “Top 10 prefecturas” and “Trayectorias de casos confirmados por Prefectura” should be “Trayectorias de casos confirmados por prefectura”

liquidx:
The fixes are now applied, thanks!

PR #377, opened 2020-07-06
Fix small Spanish translation errors.

vlumi:
Hi @reustle 

This patch updates the "Activo" title to plural, and changes "prefectura(s)" not in the beginning of sentence into lower case, mentioned in #359.


liquidx:
Thanks for the fixes!

PR #380, opened 2020-07-08
Add the today/yesterday increments on the total row of the prefecture table

vlumi:
Hi @reustle

The lack of today's and yesterday's increment on the total row of the prefecture table felt like a sore spot, so I thought of adding them. While today's total increment is available in the KPI, you'd need to dig into "Daily Confirmed Cases" to get yesterday's increment.

Screen Shot 2020-07-08 at 17 42 15


liquidx:
Thanks, that's a nice addition. Merged your PR.

PR #393, opened 2020-07-18
Add Finnish translations for latest i18n keys.

vlumi:
Hi @reustle 

Added Finnish translations for the latest strings.


liquidx:
Thanks!

PR #394, opened 2020-07-18
Adding region's names in Arabic translation

mftatfcyclone:
I have added the region names in Arabic and changed some phrases to provide a better understanding.
I recommend adding a variable for the translation of "Most Active Regions"
@reustle @liquidx


mftatfcyclone:
For your information, Arabic language is read from right to left.

liquidx:
Thanks for the PR. I'll merge this in, there may be some other string changes as we continue to iterate on the design.

I'm aware of the Right to Left for Arabic, but unfortunately, I'm still struggling with laying out the site even for Left to Right. Maybe if I get that to a better state I'd try see if I can update it to RTL.

liquidx:
Thanks again for the translations!

PR #395, opened 2020-07-18
Fix prefecture secondary sorting within region

vlumi:
Hi @reustle @liquidx 

I noticed that the sorting of the prefectures within regions seemed to be all over the place in some regions, but it was just that the secondary sorting by active cases was in ascending order. I reversed the order, and also added a third tiebreaker by total confirmed cases, which should make the list order reasonably deterministic...

Before:
Screen Shot 2020-07-18 at 23 46 09

After:
Screen Shot 2020-07-18 at 23 50 30


liquidx:
Good point, I was trying to be a little too complicated trying pushing up "recent active" prefectures.

I think this sorting you proposed is a little more understandable.

PR #398, opened 2020-07-18
Add back deceased and confirmed numbers to desktop view.

liquidx:
Renamed confirmed to total for prefectures.
Trying to move the styling CSS blocks around to get to a more "maintainable state"


liquidx:
Attempting to add back some more detail on the desktop view.
Screen Shot 2020-07-19 at 0 15 54

vlumi:
There are a lot of numbers, and with sort order/heat map fixed to active cases and/or recent change, it's getting a little hard to understand. Could the whole thing be wrapped behind an active/deaths/total toggle, showing (and sorting by) one parameter at a time?

Issue #396, opened 2020-07-18
Add credit to @vlumi as a contrributor


liquidx:
@vlumi Thanks for all the help on the site. It would be nice to credit you on the site, would you like to submit a PR with your chosen avatar, link and name?

vlumi:
Thanks!

PR #399, opened 2020-07-18
Add avatar for vlumi.

vlumi:
Hi @reustle @liquidx 

Included my avatar #396


liquidx:
Brilliant, thanks for helping!

PR #400, opened 2020-07-18
Fixed the design when the prefecture map is not displayed.

muro3r:
Hi @reustle.

I found collapse when `prefecture-map-container` is not shown.
Occurs when the browser does not support `Mapbox GL JS`.
before

I fixed it by changing the CSS height specification to min-height
after


liquidx:
Thanks for fixing this!

PR #401, opened 2020-07-18
Expand collapse prefecture v2

dshlass:
@reustle @liquidx 
New branch and PR to add the same functionality to the newly visualized prefecture regions. I left the first three regions visible and hid the rest by default. This can be updated easily by a constant at the top of RegionalCharts.js.

ezgif com-optimize (1)


liquidx:
Great work putting together the ability to expand and collapse, especially with the smooth transitions in between.

However, comparing this with just scrolling down to see the less active regions and prefectures, it makes browsing the data less straightforward. We received feedback that some people wanted to see all the data by just scrolling up and down the page, especially on mobile. Hiding the data behind the accordion may cause some more of that issues.

Nevertheless, thank you for making an attempt at this and showing us!

PR #402, opened 2020-07-19
Switch to using active cases instead of confirmed cases.

liquidx:
Switch to using active cases rather than confirmed cases. This brings the numbers down and shows where the active areas are, and where there isn't much infections happening.

Screen Shot 2020-07-19 at 14 55 18


birdjumper:
It appears that with this move, some detail has been lost, i.e. "deaths" and "recovered" in each local area. Is there a reason for this information being omitted? We now only have Deaths and Recovered for the whole country ...

liquidx:
https://github.com/reustle/covid19japan/pull/398

PR #405, opened 2020-07-19
Updated Italian localization

JSerroni:
IT localization updated on a few lines and added on missing ones. 
Please note that Tokyo with the "i" is for the Spanish language, in the Italian language is with Y.

Thanks as always @reustle !

EDIT: Could not find the string "Most Active Regions" the proper translation in Italian would be "Regioni con più attività"


liquidx:
Thanks for the corrections!

PR #406, opened 2020-07-21
Translate region-related string in Finnish.

vlumi:
Hi @reustle @liquidx 

Finnish translations for the latest regions-related strings.


liquidx:
Thanks!

PR #407, opened 2020-07-21
Add Confirmed Case Trajectories by Region

vlumi:
Hi @reustle @liquidx 

Since the data is now grouped by regions, too, I thought the confirmed case trajectories chart being only for prefectures was lacking, so I duplicated it for regions, too:

Screen Shot 2020-07-21 at 14 26 23


liquidx:
I like this idea of plotting it by region to see. Could you stop plotting the prefecture one too? I don't think we need to plot both.

vlumi:
I agree, there's a lot of redundancy in showing the prefectures alongside regions. I updated the PR with the prefecture chart removed, and the region chart made the same width as the content above:

Screen Shot 2020-07-22 at 8 47 39

liquidx:
Thanks!

PR #412, opened 2020-07-25
Translation of phrase "confirmed-case-trajectories-by-region" in Arabic

mftatfcyclone:
The traslation for "confirmed-case-trajectories-by-region" has been updated
@liquidx


liquidx:
Thanks for the updated translations

PR #413, opened 2020-07-25
Ukrainian localization for regions and grammar fixes

nagornyi:
Localized the names of regions and improved the readability of existing captions.
@reustle


liquidx:
Thanks for the updated translations!

PR #414, opened 2020-07-25
Protect against null's appearing in active.

liquidx:


liquidx:
Fixes some errors we got when we lost our recovered numbers.

PR #416, opened 2020-07-26
Added requirements.txt

benji011:
## Motivation
Installing with `pipenv` was weird for me. So maybe having another option with `requirements.txt` is also helpful, i feel.

## Changes
- Created `requirements.txt` by using `pip freeze` from the root dir in the backend dir
- Added `requirements.txt` into the backend. You can then install dependencies like below before doing migrations then running the server like:

```bash
pip3 install -r requirements.txt
python3 manage.py makemigrations
python3 manage.py migrate
python3 manage.py runserver 8000
```

## Additional notes
Hi @reustle 

I'm interested in helping out on the backend from time to time because I work with Django daily and i think you guys you use an extra hand. I'll help create the DRF side for you if you need.


reustle:
Hey @benji011 

Thanks for the PR! This is a very old and unmaintained branch. None of this code is used in production. Did you find something in it useful for your own project(s)? Thanks

benji011:
Ah i see, sorry i didn't really check properly on where the data was being imported from (which I found now is coming from [this repo](https://github.com/reustle/covid19japan-data/) from [this spreadsheet](https://docs.google.com/spreadsheets/d/1vkw_Lku7F_F3F_iNmFFrDq9j7-tQ6EmZPOLpLt-s3TY/edit#gid=434907293))

If we don't need this backend then i guess these changes can be disregarded :)

liquidx:
I'll close this for now.

PR #418, opened 2020-07-27
Replacing the KPI section with React

dshlass:
@reustle @liquidx 
#237 
I wanted to give this a try and test out the performance of the page locally. I'd love to continue the discussion of migrating away from vanilla in favour of React or even NextJS for it's built-in server side rendering. 

Contents of this PR:
- added react, react-dom, and babel-react to start using React in this repo.
- added and configured eslint to include react and react-hooks.
- added react-i18next to allow for i18n hook to get current language and set translations within React.
- created a container component (KpiReact > index.js) which takes in the ddb data and constructs the kpi data to map over.
- created a presentational component (KpiReact > kpi.js) which fetches each individual graph and renders the kpi data.
- included a helper (KpiReact > linearGradient.js) to add the gradient overlay of the svg charts

I am aware that there is some layout jump at the beginning during the render phase. If anyone has a suggestion or recommendation to reduce that please let me know.

I've attached lighthouse reports for before and after the addition of React.
[lighthouse.zip](https://github.com/reustle/covid19japan/files/4978993/lighthouse.zip)

**Summary:** 
**Speed Index:** 1.0s to 0.5s
**Time to Interactive:** 8.4s to 6.5s
**Total Blocking Time:** 5,810ms to 4,970ms
**Cumulative Layout Shift:** 0.03s to 0.014s

**No visual change:**
Screen Shot 2020-07-27 at 3 59 43 PM

Thanks for reading!


liquidx:
This is interesting, I'm not that familiar with React. One thing I noticed is that now, when the page first loads and the data is not yet loaded, the KPI boxes are not there.

Is there a way that has the original loading state to stay like it does in the current site?

dshlass:
> This is interesting, I'm not that familiar with React. One thing I noticed is that now, when the page first loads and the data is not yet loaded, the KPI boxes are not there.
> 
> Is there a way that has the original loading state to stay like it does in the current site?

There is, but I am having some difficulty implementing it without a package - which I am hesitant to add keep the packages lean. 

I'd love some input on methods to render the kpi children to specific html elements rather than a parent node (render each KPI section individually to the target section in the HTML code). 

One solution I was looking into was to render the component at the start of the "DOMContentLoaded" event without data and use redux or another global state management tool to keep track of the changing of the global data values. The problem is that without a discussion on the migration strategy to React it could overcomplicate things fairly quickly. 

I've never converted an app from vanilla to React in pieces and I can see that is will be a fun challenge. I've got lots of time to play around with the implementation, I'd just love some input from another React dev on the matter.

vlumi:
> > Is there a way that has the original loading state to stay like it does in the current site?
> 
> There is, but I am having some difficulty implementing it without a package - which I am hesitant to add keep the packages lean.

One quick hack would be to leave the original static DOM in place, so that it'll be shown until React replaces it with its own rendering. Reverting your changes to `src/index.html` seemed to work fairly well, though there is some wobbling and a quick flick when the kpi element is empty for a short moment...

dshlass:
>  Reverting your changes to `src/index.html` seemed to work fairly well, though there is some wobbling and a quick flick when the kpi element is empty for a short moment...

Yeah I actually tried this as well but thought the site _felt_ better when it was just delayed loading rather than the flick/wobble. I am going to try to set a min-height for the kpi section and see if having the blank space without the page jump feels better.

vlumi:
Splitting the rendering to each individual box makes it behave more like the current implementation, though modifying the containers' attributes now requires more code outside of the React implementation.

I made a quick prototype of this in https://github.com/dshlass/covid19japan/pull/1, but I didn't measure performance, and there seem to be some discrepancies in the layout (maybe changes in master...?). Just food for thought.

dshlass:
> Splitting the rendering to each individual box makes it behave more like the current implementation, though modifying the containers' attributes now requires more code outside of the React implementation.
> 
> I made a quick prototype of this in [dshlass#1](https://github.com/dshlass/covid19japan/pull/1), but I didn't measure performance, and there seem to be some discrepancies in the layout (maybe changes in master...?). Just food for thought.

Wow thank you for this! It feels much like the live site does. I wanted to try something similar but couldn't wrap my head about it - great solution!

I noticed the discrepancies in master as well and I think they are due to the addition of Russian and Thai to the languages.

Attached are the previous lighthouse scores with the addition of your PR (afterSections.html)
[lighthouse.zip](https://github.com/reustle/covid19japan/files/5039187/lighthouse.zip)

Here is the attached performance differences
Summary:
**Speed Index:** 1.0s to 0.5s to **0.8s**
**Time to Interactive:** 8.4s to 6.5s to **4.1s**
**Total Blocking Time:** 5,810ms to 4,970ms to **2,910ms**
**Cumulative Layout Shift:** 0.03s to 0.014s to **0.095**

These lighthouse scores are pretty inconsistent but I agree that your PR make it feel the same as the live version.

I think that as the project slowly starts to migrate towards react there will be a lot more code being added outside of the React implementation. Once it gets closer to completion most of it can be removed and it can be a functioning React app with a single ReactDOM.render call.

This is my first time contributing to an open source project. Can I merge your PR into my branch without affecting anything else? I'd love to include your code in my PR.

vlumi:
> I think that as the project slowly starts to migrate towards react there will be a lot more code being added outside of the React implementation. Once it gets closer to completion most of it can be removed and it can be a functioning React app with a single ReactDOM.render call.

I agree, the biggest turning point being when the data loading can be shifted to React, to allow the components to be drawn with default state before the data is ready.

> This is my first time contributing to an open source project. Can I merge your PR into my branch without affecting anything else? I'd love to include your code in my PR.

Of course, I created the pull request against your repo just so that there wouldn't be conflicting pull request against this main repo.

stalkerg:
Probably Svelte will be much faster and simpler in this case, same as on the NYT tracker.

src/components/KpiReact/index.js
```diff
@@ -0,0 +1,186 @@
...
+      let testedPercentage = parseInt((totals.confirmed / totals.tested) * 100);
+      if (testedPercentage < 1) {
+        testedPercentage =
+          parseInt((totals.confirmed / totals.tested) * 1000) / 10;
+      }
+      tested.percentage = t("tested-percentage", {
+        percent: formatNumber(testedPercentage),
+      });
+    }
+    setDrawTested(tested);
```
AndriiAndrey, line 144: Why do you need to put this logic inside of useEffect and store in state chart data for all types if it renders in a loop for a single type, plus if data comes outside of React and data is already loaded on the moment it mount in DOM and there is no way it will change inside of component? Aggregating of chart data for each kpi-type looks the same, it cold be moved to method to reduce repeated code for example something like this 
```
const chart = {
      id: type,
      label: t(`kpi-${type}`),
      value: formatNumber(totals[type]),
      diff:
        totalsDiff[type] >= 0
          ? `+${formatNumber(totalsDiff[type])}`
          : formatNumber(totalsDiff[type]),
      ...
```

src/components/KpiReact/index.js
```diff
@@ -0,0 +1,186 @@
...
+    case "confirmed":
+      return drawConfirmed ? draw(drawConfirmed) : "";
+    case "recovered":
+      return drawRecovered ? draw(drawRecovered) : "";
+    case "deceased":
+      return drawDeaths ? draw(drawDeaths) : "";
+    case "active":
+      return drawActive ? draw(drawActive) : "";
+    case "tested":
+      return drawTested ? draw(drawTested) : "";
```
AndriiAndrey, line 171: If you move aggregating of chart data in a separate method you can get rid of this switch case and render Kpi component with correct data depends on a type

src/components/KpiReact/index.js
```diff
@@ -0,0 +1,186 @@
...
+
+    default:
+      return "";
+  }
+};
+
+KpiContainer.propTypes = {
+  data: Proptypes.object,
+};
```
AndriiAndrey, line 180: missing Proptype for type prop

src/components/KpiReact/kpi.js
```diff
@@ -0,0 +1,91 @@
...
+        <div
+          className="chart"
+          id={`kpi-${id}-chart`}
+          dangerouslySetInnerHTML={{
+            __html: `${gradient[id]}${graph}`,
+          }}
+        ></div>
+      )}
+      {!isLoading && (
+        <p
```
AndriiAndrey, line 58: Looks like it have some issues for Tested kpi chart, caption and chart svg should be on a different lines

src/index.js
```diff
@@ -55,6 +55,9 @@ import {
 import travelRestrictions from "./data/travelRestrictions"; // refer to the keys under "countries" in the i18n files for names
 import { LANGUAGES, LANGUAGE_NAMES } from "./i18n";
 
+import React from "react";
+import ReactDOM from "react-dom";
```
AndriiAndrey, line 59: could you organise imports, put 
```
import React from "react";
import ReactDOM from "react-dom";
```
on top to other module imports

AndriiAndrey (changes requested):

AndriiAndrey:
Hi @dshlass it looks like this PR opened long time ago and it is already outdated, could you please update your repo and pull master brunch into it and solve conflicts, thanks

PR #423, opened 2020-07-28
Improved the German localization

NicoleRauch:
@reustle Here are some improvements for the German localization:

- Improved two existing translations (I found them to be a bit unusual)
- Corrected one existing translation
- Added missing translations


vlumi:
The key "languages" still seems to miss its German translation, as mentioned in #351, if it would be possible to get here, too...

NicoleRauch:
@vlumi Totally overlooked that 🤦‍♀️ Thanks for pointing it out! Amended the pull request.

NicoleRauch:
BTW this PR also fixes an issue with text overflow that the German version of the page is currently facing:

image

My proposed text is shorter and does not overflow any more.

liquidx:
Thanks, this looks great! I'll merge this in

PR #424, opened 2020-07-30
Update es.json

AlanYokohama:
I also want to propose Spanish translations for these two, but I cannot find them in the i18n file:

Most Active Regions --- Regiones más afectadas
Other --- Otras

I will be available for any other Spanish translation.


vlumi:
@AlanYokohama 

You can always compare to `en.json` for any missing keys from the language file, as they are often not added to all files when new strings are added. Best to try to add to the same position so that comparing is easier later.

The two ones you mentioned are:
```
"active-regions": "Most Active Regions",
```
```
  "regions": {
    "Other": "Other"
  },
```

AlanYokohama:
> 
> 
> @AlanYokohama
> 
> You can always compare to `en.json` for any missing keys from the language file, as they are often not added to all files when new strings are added. Best to try to add to the same position so that comparing is easier later.
> 
> The two ones you mentioned are:
> 
> ```
> "active-regions": "Most Active Regions",
> ```
> 
> ```
>   "regions": {
>     "Other": "Other"
>   },
> ```

Thanks, @vlumi 
I just don't know how to do that...

vlumi:
> I just don't know how to do that...

If you know which English term you want to translate, you could just look it up (e.g. "Most Active Regions") in `en.json`, copy-paste the whole row to `es.json` and translate the value (but keep the key as-is). If a key doesn't exist in the selected language, it'll use the English value by default.

AlanYokohama:
Thanks, @vlumi 
I just tried what you said. I hope it's the right way.
I will look forward to your news.

liquidx:
Looks like you did it right, thanks for suggesting the changes. I don't speak Spanish, so I'll trust you on these 👍

PR #432, opened 2020-08-03
Add Thai translations

pohr:
@reustle 
Add Thai translations to the site.


pohr:
Since this PR might overflow the languages to the second row.
So I also propose to reduce the font size of the languages bar to 12px to keep all languages on the same row.

https://github.com/reustle/covid19japan/pull/433

Please Check

liquidx:
Thanks for contributing the Thai translations.

PR #433, opened 2020-08-03
Reduce the font size of languages bar

pohr:
Reduce the font size of the languages bar to 12px to keep all the languages on the same row.


liquidx:
Since the languages are already overflowing to the second row on the narrower layouts, there's not need to reduce the font size. Thanks for considering that though.

PR #435, opened 2020-08-05
Update th.json

pohr:
Hi @reustle @liquidx 

I did some reading and found a mistake in translation.
Here is PR to fix it.


liquidx:
Thanks for the fixes!

PR #436, opened 2020-08-11
Missing translation added for IT.json

JSerroni:
"Confirmed Case Trajectories by Region" translated in Italian with "Curva dei casi confermati per Regione". In Italian language it is differentiated between trajectory  and curve. With this kind of graphs "curva" is the proper one.

@reustle 


liquidx:
Thanks for helping with the translations!

Issue #451, opened 2020-11-20
Graph for number of tests over time 


osdiab:
// Please tag @reustle in your newly created issue
@reustle 

From the perspective of “is the government testing people” would be nice to see a similar tests over time graph like the cases one

![D4D254AD-E8E1-465F-982A-69D879C43E3E](https://user-images.githubusercontent.com/1890926/99796271-8903ef00-2b70-11eb-8448-935e14aff62a.jpeg)

liquidx:
Added an expanded graph that shows the number of tests over time. 

https://covid19japan.com/#tested

Issue #447, opened 2020-11-01
Redesign the "Total cases" graph (?)


ggouaillardet:
@reustle Currently, the "Total cases" graph (the first one on the left from the web interface) plots
 - Confirmed: ~100k
 - Recovered: ~100k
 - Active: ~10k
 - Deaths: ~1k

These are three order of magnitudes plotted on the same linear graph
 - Confirmed cases is easily readable
 - Recovered cases is also easily readable, and basically "follows" the Confirmed cases line
 - Active is harder to read
 - Deaths is basically flat and impossible to read.

There are imho different options to improve the readability:
 1. plot on a logarithmic scale
 2. Stop plotting Confirmed and Recovered lines
 3. plot the new confirmed cases (~1k, similar range than Deaths)

liquidx:
I agree that the total cases chart is getting a little useless now. 

I think we should plot just the active cases and then a separate one for deaths?

reustle:
Thanks for submitting this, and I agree with @liquidx . I think a separate graph would be a good move. I appreciate it!

liquidx:
Merged #454 that splits these charts up now. I've also added a way to look at the last 3 months vs 6 months (right now labelled all-time).

PR #454, opened 2020-11-21
Split up trends charts into different charts

liquidx:
Attempt to address #451 and #447 

Refactor DailyIncrease component into something that can be reused for the Active, Deceased and Tested charts. 
Replace total trends chart with the DailyIncrease charts showing Active, Deceased and Tested.
Default to showing 6 months in the large charts, add option to show 3 months or 6 months (all time).
Fix rendering artifacts when showing 6 months of data.
Comment out the initial chart rendering code as it doesn't seem to be called.


liquidx:
Screenshot:

Screen Shot 2020-11-21 at 23 46 54
Screen Shot 2020-11-21 at 23 47 38

src/index.js
```diff
@@ -257,6 +308,7 @@ const loadDataOnPage = () => {
       ddb.totals = newTotals[0];
       ddb.totalsDiff = newTotals[1];
       ddb.trend = summaryData.daily;
+      console.log(ddb.trend);
```
reustle, line 311: Maybe we can get rid of this when we deploy?

src/index.js
```diff
@@ -168,18 +174,33 @@ const setLang = (lng) => {
...
-          CHART_TIME_PERIOD
-        );
-        dailyIncreaseChart = drawDailyIncreaseChart(
-          ddb.trend,
-          dailyIncreaseChart,
-          LANG,
-          CHART_TIME_PERIOD
-        );
+
+        // TODO: This doesn't seem to be needed and can be removed.
```
reustle, line 178: Maybe it's safe to remove this too

reustle:
This looks great! Thanks for putting this together, it's definitely something that people have been looking forward to seeing.

src/index.js
liquidx: Good catch, fixing.

src/index.js
liquidx: I'm thinking of removing this in a separate change, I'm trying to figure out if there's a corner case we used to have which we kept this around for.

Issue #449, opened 2020-11-13
Link to countries' restrictions PDF is broken


NicoleRauch:
@reustle The link http://www.moj.go.jp/content/001329914.pdf does not work any more...

BTW it seems that all countries link to the same PDF file; isn't it a bit wasteful to add so many links (and make the user search for the desired country in this dense list, and making them - or at least me - believe that they will immediately get information specifically tailored to the country)? In other words, wouldn't a single link suffice?

liquidx:
I agree here that it's not that useful to link all the countries any more.

Maybe better to point visitors to either the rules for Foreign Resident Re-Entry:
http://www.moj.go.jp/nyuukokukanri/kouhou/nyuukokukanri07_00245.html

There is also a link on that site to the probably updated PDF (in Japanese) that describes the full list:
http://www.moj.go.jp/content/001327502.pdf

reustle:
I also agree, thanks for posting this. Maybe we could remove this section entirely and just add a link to the page(s) that @liquidx provided?

NicoleRauch:
The link to the PDF seems to be quite fragile (looking at the past issues, it already changed at least once before) and it's prominently linked in the website behind the first page, so my vote would be for just adding that page and letting the users access the PDF from there.

liquidx:
Removed the PDFs and now linked to a variety of government and news sites that describe the latest rules.

Issue #277, opened 2020-04-29
Cleanup Travel Restrictions


reustle:
We can hide or remove these currently unused sections

![image](https://user-images.githubusercontent.com/304560/80586947-b7916f80-8a50-11ea-9a9e-7688e465d901.png)

ekerik220:
https://github.com/reustle/covid19japan/pull/278

This will just hide the sections when they are empty. Just in these sections are used again in the future.

mertd:
Are there any other changes necessary other than #278 ?

liquidx:
Removed those unused sections now.

Issue #161, opened 2020-03-31
Add anchor links to H1 headings


ann-kilzer:
Now that the site has many more informative sections, it would be helpful to have some anchor links so that we can share specific parts of the site, e.g. `https://covid19japan.com/#travel-restrictions` 

Thoughts @reustle ?

reustle:
💯 👌 


@Kay-Cube also talked about adding links soon, but I think we'll have a higher level discussion about design soon

liquidx:
After nearly 8 months 🤦🏻‍♂️, this anchor link now works.

PR #460, opened 2020-11-25
Replace the list of banned countries with links to latest information from MOJ

liquidx:
Removed the banned list of countries for travel, because technically they're not banned any more since November. The PDF that all the countries pointed to were broken, so it's not accurate information any more.

Instead, this PR replaces the information with links to the MOJ information and a recent Japan Times article.


liquidx:
Draft changes for the travel restrictions .. any suggestions @reustle ?

reustle:
Looks great! 👏 :shipit:

liquidx:
Fixes #449 #277 #161

PR #467, opened 2020-11-30
Add missing Finnish translations for the time period functionality.

vlumi:
Hi @reustle, @liquidx 

I added Finnish translations to the new translation strings added in #454.


liquidx:
Thanks for adding the new translations!

PR #468, opened 2020-11-30
Added German translation for the new strings.

NicoleRauch:
@reustle I've added German texts for the new strings.


reustle:
Thank you!

PR #470, opened 2020-12-12
Update th.json

pohr:
Hi @reustle , @liquidx 

Here's an updated translation for the Thai language.


liquidx:
Thank you for the updated translations!

PR #471, opened 2020-12-16
Update es.json

AlanYokohama:
Dear @reustle 

These are the latest translations for Spanish.


liquidx:
Thank you for the updated translations!

PR #477, opened 2021-01-18
Small changes and loc updated to match last design

JSerroni:
Added: language, prefectures and regions list, translated line 127 “other” for regions. Please note that "Tokyo" string is missing under prefectures, because now appears in alphabetical order. 

Please updated @reustle :)


liquidx:
Thanks for the PR with the fixes!

PR #487, opened 2021-04-22
Replace language picker section with dropdown

AndriiAndrey:
Replaced language picker section with dropdown component using react

Screenshot 2021-04-22 at 14 28 24
Screenshot 2021-04-22 at 14 30 08


src/index.html
```diff
@@ -60,6 +60,7 @@ <h1 data-i18n="covid-19-tracker">Japan COVID-19 Coronavirus Tracker</h1>
         <div class="lang-picker">
           <span class="lang-picker-title" data-i18n="languages">Languages:</span>
```
reustle, line 61: Since it will now be a dropdown, it should be changed to `Language`. I'm not terribly worried about changing it from plural to singular in every language right now.

reustle (approved):

src/index.html
AndriiAndrey: Fixed

PR #499, opened 2021-08-01
Update Japanese localization

shuuji3:
Hi, I've updated the translation of three texts into Japanese.

**Screenshots**

Screen Shot 2021-08-02 at 0 02 17
Screen Shot 2021-08-02 at 0 02 30


liquidx:
Thank you so much for the updated translations.

shuuji3:
Thank you! 🙂

PR #508, opened 2022-01-23
Fix dead link

n194:
The original link goes returns a 404. The new link has an updated version of the original guide.


reustle:
Thank you!

PR #516, opened 2022-08-04
Fix number overlap

alvihrhp:

1. Add new breakpoint variable 375px and 425px on _layout.scss,
2. Add new breakpoint 375px on _regional.scss to decrease the font-size by .4 rem and the diff display to inline-block


reustle:
Nice work, thanks @alvihrhp !

